PR #19, opened 2025-02-07
add cramino to the main lr_somatic.nf workflow

AlexanRNA:
adding cramino pre and post alignment to the workflow

### DIFF
--- a/workflows/lr_somatic.nf
+++ b/workflows/lr_somatic.nf
@@ -16,6 +16,7 @@ include { SAMTOOLS_CAT        } from '../modules/nf-core/samtools/cat/main'
 include { MINIMAP2_INDEX      } from '../modules/nf-core/minimap2/index/main'
 include { CLAIRSTO            } from '../modules/local/clairsto/main'
 include { MINIMAP2_ALIGN      } from '../modules/nf-core/minimap2/align/main'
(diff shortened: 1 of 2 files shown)


workflows/lr_somatic.nf
```diff
@@ -64,9 +65,9 @@ workflow LR_SOMATIC {
     // Maybe add a subworkflow with all pre-alignment QC together if there will be more than CRAMINO
     //
     // MODULE: CRAMINO
-    //
-    CRAMINO_PRE ( )
     */
+    CRAMINO_PRE ( ch_cat_ubams )
+
```
ljwharbers, line 70: You should add the versions channel and mix it with the rest of the versions. Just look at some of the other processes as example how.

workflows/lr_somatic.nf
```diff
@@ -119,7 +120,7 @@ workflow LR_SOMATIC {
     // MODULE: CRAMINO
     // 
     
-    //CRAMINO_POST ( )
+    CRAMINO_POST ( ch_minimap_bam )
```
ljwharbers, line 124: Same for this one, also add versions here (doesnt matter that its the same version as above)

ljwharbers (changes requested):

AlexanRNA:
@ljwharbers the version channels have been added

ljwharbers (approved):

PR #189, opened 2026-09-03
Add savana module

YannVRB:
Summary

  Adds SAVANA structural variant and copy-number calling as a second SV/CNA caller alongside Severus/ASCAT, following the existing paired vs tumor-only subworkflow split.

  - SAVANA_RUN → SAVANA_CLASSIFY → SAVANA_CNA for paired samples; SAVANA_TO (bundled CNA) for tumor-only.
  - Inputs are the haplotagged BAMs from PHASING_HAPLOTYPING, per SAVANA's own recommendation, with --contigs restriction and --g1000_vcf population SNPs (tumor-only) wired in.
  - VEP_SAVANA annotates SAVANA's output the same way SV_VEP annotates Severus's.
  - Vendored savana/run and savana/classify modules are patched (tracked via .diff) to add contig-file support; container bumped to 1.3.8.

  Review
  
  Addressed all of @ljwharbers's review feedback:
  - Fixed the one-sample-per-run bug (fasta.combine(fai...) collapsing to a single-shot queue channel via .first()).
  - Fixed PacBio samples being classified with the ONT model.
  - Fixed CNA consuming raw breakpoints instead of the classified somatic VCF.
  - Switched to haplotagged BAMs, added --contigs, and tumor-only now uses --g1000_vcf instead of the tumor's own germline calls.
  - Updated tests/default.nf.test.snap and .nftignore, added a test_chm13 profile.
  - Ran pre-commit/nf-core lint, updated CITATIONS.md/README.md/CHANGELOG.md/docs, trimmed comments.
  - Pushed back (with SAVANA source citations) on dropping --chromosomes 19 in the test configs — confirmed it's a correct 1-based index into SAVANA's hardcoded chromosome list, not an out-of-range bug,
    and that removing it causes a real crash on partial references. Detail in the review thread.

  Testing

  - -profile test and -profile test_chm13: both pass end-to-end, snapshot regenerated.
  - Validated on real data (not just CI's sparse single-contig test data): COLO829 (public) and an internal clinical GBM sample, subset to chr5+chr19 for speed. Full SAVANA_RUN → CLASSIFY → CNA chain
    completes with a real purity/ploidy fit and a properly segmented copy-number profile backed by thousands of het-SNPs per segment.
  - Incidentally found and fixed (separately, in #191) a pre-existing Clair3 model-staging bug hit during real-data testing — unrelated to SAVANA, kept out of this PR.

PR checklist

  - [x] This comment contains a description of changes (with reason).
  - [x] If you've fixed a bug or added code that should be tested, add tests!
  - [x] If you've added a new tool - have you followed the pipeline conventions in the contribution docs (https://github.com/IntGenomicsLab/lrsomatic/tree/main/docs/CONTRIBUTING.md)
  - [x] Make sure your code lints (nf-core pipelines lint).
  - [x] Ensure the test suite passes (nextflow run . -profile test,docker --outdir <OUTDIR>).
  - [ ] Check for unexpected warnings in debug mode (nextflow run . -profile debug,test,docker --outdir <OUTDIR>).
  - [x] Usage Documentation in docs/usage.md is updated.
  - [x] Output Documentation in docs/output.md is updated.
  - [x] CHANGELOG.md is updated.
  - [x] README.md is updated (including new tool citations and authors/contributors).


subworkflows/local/paired/paired_savana.nf
```diff
@@ -0,0 +1,76 @@
...
+
+workflow PAIRED_SAVANA {
+
+    take:
+    tumor_normal_input // [meta, tumor_bam, tumor_bai, normal_bam, normal_bai, phased_germline_vcf, phased_germline_tbi]
+    fasta              // [[:], fasta]
+    fai                // [[:], fai]
+
+    main:
+    ch_fasta_fai = fasta.combine(fai.map { it[1] })
```
ljwharbers: **Bug: SAVANA only runs for one sample.** `fasta.combine(fai.map { it[1] })` on two value channels yields a *queue* channel with a single item, so `SAVANA_RUN`/`SAVANA_CNA` fire once and the other samples silently get no SAVANA output. Verified with a minimal Nextflow 26.04.3 script (value channel passed directly: 3 tasks; `combine` result: 1 task). Add `.first()` here (and in `tumoronly_savana.nf`), or build the `[meta, fasta, fai]` tuple once in the main workflow.

Also: `nextflow lint` warns on the implicit `it`; use an explicit closure parameter.

subworkflows/local/tumor_only/tumoronly_savana.nf
```diff
@@ -0,0 +1,52 @@
...
+
+workflow TUMORONLY_SAVANA {
+
+    take:
+    tumor_input // [meta, tumor_bam, tumor_bai, phased_germline_vcf, phased_germline_tbi]  -- tumor-only, no matched normal
+    fasta       // [[:], fasta]
+    fai         // [[:], fai]
+
+    main:
+    ch_fasta_fai = fasta.combine(fai.map { it[1] })
```
ljwharbers: Same one-sample bug as in `paired_savana.nf`: this needs `.first()` (or a value-channel tuple built upstream), otherwise `SAVANA_TO` runs for only one tumor-only sample.

subworkflows/local/paired/paired_savana.nf
```diff
@@ -0,0 +1,76 @@
...
+    //                                 SAVANA-native TSV this pipeline's downstream consumer already
+    //                                 parses (chromosome,start,end,segment_id,bin_count,
+    //                                 sum_of_bin_lengths,weight,copyNumber[,...] -- columns 1,2,3,8).
+    //
+    tumor_normal_input
+        .map { meta, tumor_bam, tumor_bai, normal_bam, normal_bai, phased_vcf, _phased_tbi ->
+            def allele_counts_het_snps = []
+            return [meta, tumor_bam, tumor_bai, normal_bam, normal_bai, phased_vcf, allele_counts_het_snps]
+        }
+        .join(SAVANA_RUN.out.sv_breakpoints_vcf)
```
ljwharbers: **Wrong `--breakpoints` input for CNA.** SAVANA's own combined workflow (`savana_main` in savana.py) hands the *classified somatic* VCF to the CNA step (`args.breakpoints = args.somatic_output`). Here the raw, unclassified `sv_breakpoints_vcf` is used, which still contains germline and artefact junctions. Join on `SAVANA_CLASSIFY.out.somatic_vcf` instead (note it is `optional: true` in the module, so tolerate a missing file or fall back to `[]`).

subworkflows/local/paired/paired_savana.nf
```diff
@@ -0,0 +1,76 @@
...
+    // MODULE: SAVANA_CNA (label: process_high)
+    // Input:  [meta, tumor_bam, tumor_bai, normal_bam, normal_bai, snp_vcf, [], breakpoints]
+    //           snp_vcf     -- reuses the phased germline VCF already computed for SEVERUS/phasing
+    //           breakpoints -- SAVANA_RUN's raw breakpoints (informs CNA binning around SV junctions)
+    //         [[:], fasta, fai]
+    //         [] / [] / []  -- contigs / blacklist / g1000_vcf (all unused; snp_vcf takes priority)
+    // Output: .cna -- [meta, tsv]  -- segmented absolute copy number. Same column shape as the
+    //                                 SAVANA-native TSV this pipeline's downstream consumer already
+    //                                 parses (chromosome,start,end,segment_id,bin_count,
+    //                                 sum_of_bin_lengths,weight,copyNumber[,...] -- columns 1,2,3,8).
```
ljwharbers: This describes a downstream consumer that does not exist in the repo. Please drop the comment; in general the comment blocks in this PR are far longer than the surrounding code style, one or two lines per block is enough.

subworkflows/local/paired/paired_savana.nf
```diff
@@ -0,0 +1,76 @@
+// IMPORT MODULES
+include { SAVANA_RUN      } from '../../../modules/nf-core/savana/run/main'
+include { SAVANA_CLASSIFY } from '../../../modules/nf-core/savana/classify/main'
+include { SAVANA_CNA      } from '../../../modules/nf-core/savana/cna/main'
+
+workflow PAIRED_SAVANA {
+
+    take:
+    tumor_normal_input // [meta, tumor_bam, tumor_bai, normal_bam, normal_bai, phased_germline_vcf, phased_germline_tbi]
```
ljwharbers, line 9: **Use the haplotagged BAMs.** SAVANA reads the `HP` tag per read and the README recommends phased tumour and normal BAMs. `PHASING_HAPLOTYPING` already emits `tumor_normal_hapbams_ch` for every sample, but this PR feeds the unphased minimap BAMs from `severus_input`. Consider building the SAVANA input from the haplotagged BAM channel.

subworkflows/local/tumor_only/tumoronly_savana.nf
```diff
@@ -0,0 +1,52 @@
...
+        }
+        .set { savana_to_input }
+    // savana_to_input: [meta, tumor_bam, tumor_bai, snp_vcf, [], []]
+
+    SAVANA_TO (
+        savana_to_input,
+        ch_fasta_fai,
+        [[:], []],  // contigs (unused)
+        [[:], []],  // blacklist (unused)
+        [[:], []]   // g1000_vcf (unused -- snp_vcf takes priority)
```
ljwharbers: For tumor-only, SAVANA's README recommends the bundled population SNPs via `--g1000_vcf` (`1000g_hg38` / `1000g_t2t`, so CHM13 is covered) rather than a SNP VCF. Using the tumour's own germline call set as `--snp_vcf` pulls somatic SNVs and LOH sites into the het-SNP set used for purity fitting. Suggest `g1000_vcf = params.genome == 'CHM13' ? '1000g_t2t' : '1000g_hg38'` here and keep `--snp_vcf` for paired mode only.

The README also strongly recommends gnomAD-SV / HMF PoN filtering of tumor-only SVs; fine as a follow-up, but worth a note in the docs.

workflows/lrsomatic.nf
```diff
@@ -941,6 +943,58 @@ workflow LRSOMATIC {
...
+    // subworkflow variants rather than one shared process.
+    // Input:  severus_input, branched on whether normal_bam is present
+    //         ch_fasta / ch_fai
+    // Output: .somatic_vcf -- [meta, vcf]  -- classified somatic SV VCF
+    //         .cn_calls    -- [meta, tsv]  -- segmented absolute copy number
+    //
+
+    if (!params.skip_savana) {
+        severus_input
+            .branch { meta, tumor_bam, tumor_bai, normal_bam, normal_bai, phased_vcf, phased_tbi ->
```
ljwharbers: Six unused closure parameters here; prefix them with `_` to silence `nextflow lint`.

workflows/lrsomatic.nf
```diff
@@ -941,6 +943,58 @@ workflow LRSOMATIC {
...
+
+        PAIRED_SAVANA (
+            branched_savana_input.paired,
+            ch_fasta,
+            ch_fai
+        )
+
+        TUMORONLY_SAVANA.out.somatic_vcf
+            .mix(PAIRED_SAVANA.out.somatic_vcf)
+            .set { savana_somatic_vcf }
```
ljwharbers, line 1039: `savana_somatic_vcf` and `savana_cn_calls` (below) are set but never consumed (`nextflow lint`: "Variable was declared but not used"). Either wire them into something (VEP, report) or drop these two blocks and the subworkflow emits.

workflows/lrsomatic.nf
```diff
@@ -941,6 +943,58 @@ workflow LRSOMATIC {
         ch_versions = ch_versions.mix(ASCAT.out.versions)
     }
 
+    //
```
ljwharbers, line 949: Please shorten this to two lines: what runs, and that tumor-only uses `savana to` while paired uses `run` + `classify` + `cna`.

conf/modules.config
```diff
@@ -431,6 +431,14 @@ process {
         ]
     }
 
+    withName: '.*:SAVANA_.*' {
```
ljwharbers, line 434: **Missing platform switch.** `savana classify`/`savana to` default to the ONT model when neither `--ont` nor `--pb` is given (savana.py classify setup), so PacBio samples get ONT classification. Every other caller in this file switches on `meta.platform == 'pb'` (lines 172-196, 292-296); add the same for `.*:SAVANA_CLASSIFY` and `.*:SAVANA_TO`, e.g. `ext.args = { meta.platform == 'pb' ? '--pb' : '--ont' }`. The README also suggests a higher `--min_support` for PacBio (10 worked best for them); worth exposing as a param.

**Contigs.** The README recommends `--contigs` (one contig per line, e.g. `example/contigs.chr.hg38.txt` = chr1-22, X, Y) so alt/decoy/unplaced contigs are skipped. WAKHAN already gets a contigs argument in this file (line 550); `SAVANA_RUN`/`SAVANA_TO` should get a genome-aware equivalent (a file per genome in `conf/igenomes.config` would be cleanest). Only for `run`/`to`; `cna` does not take `--contigs`.

Minor: everything lands under `variants/savana`, mixing SV and CN output while ASCAT's CN lives elsewhere. Worth deciding the layout now.

conf/test.config
```diff
@@ -40,6 +40,25 @@ process {
...
+    // SAVANA's cellularity-fitting step (savana cna, run standalone or internally by savana to)
+    // can raise an unhandled StatisticsError on minimal smoke-test data -- too few real
+    // heterozygous SNPs for the allele-count-based estimator. The vendored nf-core module's own
+    // test suite already marks this exact output optional for the same reason (see
+    // modules/nf-core/savana/cna/main.nf's "not possible with test data" comments). Ignoring
+    // failures here only in the test profile keeps the rest of the pipeline (and the CI
+    // assertions, which don't depend on CNA's output) green; it does not affect real runs.
+    withName: '.*SAVANA_CNA' {
+       ext.args = {
+            "--chromosomes 19"
```
ljwharbers: **`--chromosomes 19` is wrong.** SAVANA's `-c/--chromosomes` is a 1-based index into the reference contig list (`chr_names = [contigs[int(x)-1] ...]`, bin_generator.py:142), not a chromosome name. The test reference has only chr19, so index 19 is an IndexError. This is very likely the actual crash behind `errorStrategy = 'ignore'`. Drop the argument (default `all` already means only chr19 here), rerun, and keep `errorStrategy = 'ignore'` only if a genuine `StatisticsError` remains, and then only on `SAVANA_CNA`.

ljwharbers (commented):
Thanks for the PR. The structure is right (nf-core modules with a recorded patch, topic-channel versions, paired vs tumor-only split mirroring the small-variant subworkflows, `--skip_savana` wired everywhere), but there are a few functional problems that a green run would not reveal, plus some test and hygiene items. Details inline; summary:

**Blocking**
- SAVANA runs for only one sample per pipeline run (`fasta.combine(fai...)` yields a single-shot queue channel).
- PacBio samples are classified with the ONT model (no `--pb`).
- `--chromosomes 19` in the test configs is an out-of-range contig index on the chr19-only test reference; this is most likely the crash that `errorStrategy = 'ignore'` hides.
- The CNA step gets the raw breakpoints instead of the classified somatic VCF that SAVANA's own workflow uses.

**Against SAVANA's recommendations**
- Unphased BAMs are used although haplotagged BAMs are already produced by `PHASING_HAPLOTYPING`.
- No `--contigs` restriction.
- Tumor-only uses the tumour's own germline calls as `--snp_vcf` instead of the bundled `--g1000_vcf` population set.

**Tests**
- `tests/default.nf.test.snap` is not updated; the default test snapshots every output path, so the new `*/variants/savana/*` files will change it. Please update the snapshot and add `tests/.nftignore` entries for non-deterministic SAVANA files.
- `test_chm13` samplesheet 404s until test-datasets PR #3 merges.
- CI has not run yet (fork PR awaiting approval); the red `template_version` check is expected.

**Hygiene**
- `pre-commit run --all-files` was not run (`modules.json` churn).
- `nextflow lint`: two dead variables, six unused closure params, implicit `it`.
- CITATIONS.md and README.md (tool list, output tree, contributors) need SAVANA entries; CHANGELOG has `#XXX`; PR description/checklist are empty.
- Comments are much longer than the surrounding code style, and a few describe things that don't exist (a downstream CNA consumer, a SAVANA CHM13 branch). Please trim.

conf/test.config
```diff
@@ -40,6 +40,25 @@ process {
...
+       }
+       errorStrategy = 'ignore'
+    }
+    withName: '.*SAVANA_TO' {
+       ext.args = {
+            "--chromosomes 19"
+       }
+       errorStrategy = 'ignore'
```
ljwharbers, line 54: With `errorStrategy = 'ignore'` on `SAVANA_TO`, tumor-only SAVANA is completely untested (tests/chm13.nf.test explicitly skips sample3 for this reason). Please remove once the `--chromosomes` issue above is fixed.

The 7-line justification comment is duplicated verbatim in `test_chm13.config`; one short line is enough.

conf/test_chm13.config
```diff
@@ -0,0 +1,84 @@
+/*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    Nextflow config file for running minimal tests against the CHM13 reference genome
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    Same scope as the `test` profile (same 3 samples, same skipped steps), but the samples
+    are aligned to CHM13 chr19 instead of GRCh38 chr19 and --genome CHM13 is set, so every
+    genome-conditional code path (ClairSTO/DeepSomatic PON selection, Severus PON/bed,
+    SAVANA) exercises its CHM13 branch instead of its GRCh38 branch.
```
ljwharbers: The SAVANA wiring has no genome-conditional code, so it does not "exercise a CHM13 branch". Please correct. Also consider `includeConfig 'test.config'` and override only `input`, `fasta`, `genome` instead of duplicating the whole process block.

conf/test_chm13.config
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+}
+
+params {
+    config_profile_name        = 'CHM13 test profile'
+    config_profile_description = 'Minimal test dataset (CHM13 chr19) to check pipeline function against the CHM13 reference genome'
+
+    // Input data -- same 3 samples as the `test` profile, realigned to CHM13 chr19
+    input    = "https://raw.githubusercontent.com/IntGenomicsLab/test-datasets/main/samplesheets/samplesheet_lr-somatic_chm13.csv"
```
ljwharbers: This URL returns 404. The fixtures are in the unmerged IntGenomicsLab/test-datasets PR #3 (`test-chm13-fixtures`); that has to land first and the PR description should say so. Since `tests/chm13.nf.test` is tagged `extended`, PR CI skips it but pushes to `dev` run all tags and will fail until then.

tests/chm13.nf.test
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        then {
+            assertAll(
+                { assert workflow.success},
+                { //files exist
+                    assert file("$outputDir/sample1/variants/clair3/merge_output.vcf.gz").exists()
+                    assert file("$outputDir/sample1/variants/clairs/indel.vcf.gz").exists()
+                    assert file("$outputDir/sample1/variants/severus/somatic_SVs/severus_somatic.vcf.gz").exists()
+                    assert file("$outputDir/sample1/variants/savana/sample1.classified.somatic.vcf").exists()
```
ljwharbers, line 26: These sample1/sample2 SAVANA assertions currently fail regardless of fixtures because of the one-sample `combine` bug in the subworkflows.

tests/chm13.nf.test
```diff
@@ -0,0 +1,42 @@
...
+                    assert file("$outputDir/sample2/variants/severus/somatic_SVs/severus_somatic.vcf.gz").exists()
+                    assert file("$outputDir/sample2/variants/savana/sample2.classified.somatic.vcf").exists()
+                    assert file("$outputDir/sample3/variants/clairsto/somatic.vcf.gz").exists()
+                    // sample3 (tumor-only) is intentionally not asserted on for SAVANA: savana to
+                    // bundles SV-calling + classify + CNA into one process, and its internal CNA
+                    // step is errorStrategy='ignore'd on this minimal test data (see test.config)
+                    // -- when it fails, the whole process (including the classify-equivalent
+                    // output) produces nothing, not just the CNA-specific files. Only paired mode
+                    // (sample1/sample2) keeps RUN/CLASSIFY as separate processes unaffected by
+                    // CNA's crash, so only those are safe to assert on.
```
ljwharbers: Please reduce this to a one-liner or move it to the PR description.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## v1.2.0dev
 
+### `Added`
+
+- [#XXX](https://github.com/IntGenomicsLab/lrsomatic/pull/XXX) - Added SAVANA structural variant and copy-number calling, running alongside Severus/ASCAT (@yannvrb).
```
ljwharbers: `#XXX` -> `#189`.

modules.json
```diff
@@ -8,186 +8,285 @@
                     "ascat": {
                         "branch": "master",
                         "git_sha": "e753770db613ce014b3c4bc94f6cba443427b726",
-                        "installed_by": ["modules"],
+                        "installed_by": [
+                            "modules"
```
ljwharbers: `modules.json` is reformatted wholesale (every `installed_by` array expanded, trailing newline dropped): 149/40 line churn for four real entries. Prettier via `pre-commit run --all-files` will revert this; please reduce the diff to the four new `savana/*` entries.

modules/nf-core/savana/run/main.nf
```diff
@@ -0,0 +1,50 @@
+process SAVANA_RUN {
+    tag "${meta.id}"
+    label 'process_high'
+
+    conda "${moduleDir}/environment.yml"
+    container "${workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/savana:1.3.7--pyhdfd78af_0'
```
ljwharbers: `run` and `classify` pin savana 1.3.7 while `cna` and `to` pin 1.3.8. This matches nf-core master, but running two versions in one pipeline is avoidable; consider an nf-core patch bumping `run`/`classify` to 1.3.8 (as already done for `cna`).

docs/output.md
```diff
@@ -339,6 +339,47 @@ Present in **tumor-only** samples (no matched normal).
...
 
+#### `savana`
+
+SAVANA structural variant and copy-number calling. Runs alongside Severus/ASCAT rather than replacing
+either. Matched tumor/normal samples run `savana run` + `savana classify` + `savana cna` as separate
+steps; tumor-only samples run the combined `savana to` command instead, producing the same file set
+in one step. We strongly recommend matched tumor/normal mode for best performance -- SAVANA's own
+docs note tumor-only calling is a fallback, best combined with population/panel-of-normals filtering.
+
+```
```
ljwharbers, line 350: The tree omits the emitted `*_read_counts_*_log2r_segmented.tsv` and `*kbp_bin_ref_*.bed` files.

docs/usage.md
```diff
@@ -210,6 +211,14 @@ For structural variants, the CHM13 panel of normals is a merged panel combining
 | ---------------------- | ------------------------------------------------------------------------------------ |
 | `--severus_minsupport` | Minimum number of supporting reads required for SEVERUS to call an SV. Default = `3` |
 
+#### SAVANA Options
+
+No dedicated parameters yet -- SAVANA reuses the aligned/phased BAMs and the phased germline VCF
```
ljwharbers: After adding the platform switch, contigs handling and the tumor-only `--g1000_vcf` choice, please document them here.

ljwharbers (changes requested):

YannVRB:
> Thanks for the PR. The structure is right (nf-core modules with a recorded patch, topic-channel versions, paired vs tumor-only split mirroring the small-variant subworkflows, `--skip_savana` wired everywhere), but there are a few functional problems that a green run would not reveal, plus some test and hygiene items. Details inline; summary:
> 
> **Blocking**
> 
> * SAVANA runs for only one sample per pipeline run (`fasta.combine(fai...)` yields a single-shot queue channel).
> * PacBio samples are classified with the ONT model (no `--pb`).
> * `--chromosomes 19` in the test configs is an out-of-range contig index on the chr19-only test reference; this is most likely the crash that `errorStrategy = 'ignore'` hides.
> * The CNA step gets the raw breakpoints instead of the classified somatic VCF that SAVANA's own workflow uses.
> 
> **Against SAVANA's recommendations**
> 
> * Unphased BAMs are used although haplotagged BAMs are already produced by `PHASING_HAPLOTYPING`.
> * No `--contigs` restriction.
> * Tumor-only uses the tumour's own germline calls as `--snp_vcf` instead of the bundled `--g1000_vcf` population set.
> 
> **Tests**
> 
> * `tests/default.nf.test.snap` is not updated; the default test snapshots every output path, so the new `*/variants/savana/*` files will change it. Please update the snapshot and add `tests/.nftignore` entries for non-deterministic SAVANA files.
> * `test_chm13` samplesheet 404s until test-datasets PR #3 merges.
> * CI has not run yet (fork PR awaiting approval); the red `template_version` check is expected.
> 
> **Hygiene**
> 
> * `pre-commit run --all-files` was not run (`modules.json` churn).
> * `nextflow lint`: two dead variables, six unused closure params, implicit `it`.
> * CITATIONS.md and README.md (tool list, output tree, contributors) need SAVANA entries; CHANGELOG has `#XXX`; PR description/checklist are empty.
> * Comments are much longer than the surrounding code style, and a few describe things that don't exist (a downstream CNA consumer, a SAVANA CHM13 branch). Please trim.

Traced this through SAVANA's source rather than just testing it, since the two functions that consume --chromosomes/-c (bin_generator.py:generate_bins and allele_counter.py:MakeWindows) both build their contig list the same way:

  contigs = ['1','2','3',...,'22','X','Y']  # hardcoded 24-entry list, not from the reference
  ref_contigs = fasta.references
  if any(c.startswith('chr') for c in ref_contigs):
      contigs = [f'chr{x}' for x in contigs]
  if chromosomes != 'all':
      chr_names = [contigs[int(x)-1] for x in chromosomes]

  chromosomes is a 1-based index into that fixed 24-element human-chromosome list (chr-prefixed based on whether the reference uses chr naming) — it's not indexed into whatever contigs actually exist in the FASTA. So 19 → contigs[18] → chr19,
  which is exactly the one contig our test reference has. No IndexError.

  I also checked what happens if the arg is dropped (falling back to the default all): chr_names becomes the entire 24-chromosome list, and SAVANA then calls fasta.get_reference_length(chrom) for every one of them — including chr1, which
  doesn't exist in a chr19-only reference. I hit this exact failure mode by accident during testing (a config bug on my end briefly dropped --chromosomes for one process): KeyError: 'chr1' from pysam. So removing the argument doesn't fix
  anything here — it trades a working restriction for a guaranteed crash.

  The actual traceback behind errorStrategy='ignore', unchanged before and after this round of fixes, is:

  File ".../savana/allele_counter.py", line 297, in perform_allele_counting
      allele_counts_final = process_allele_counts(out_path_interim, het_snps, ac_window)
  ...
  FileNotFoundError: File ".../sample1_allele_counts_hetSNPs_INTERIM.bed" does not exist

  That's the het-SNP sparsity issue already documented in the nf-core module's own "not possible with test data" comments — not related to --chromosomes at all. Confirmed on two full pipeline runs (-profile test and -profile test_chm13) after
  fixing the other three blocking bugs: --chromosomes 19 stays, errorStrategy='ignore' stays, and it's scoped correctly.

  Happy to be pushed back on if I'm missing something, but I'd rather keep the working restriction than swap it for the default that's now confirmed to crash on partial references.

subworkflows/local/paired/paired_savana.nf
YannVRB: Fixed — added .first() after the combine, and replaced the implicit it[1] with an explicit closure parameter per your note below. Confirmed on two real runs that both samples now get SAVANA output.

subworkflows/local/tumor_only/tumoronly_savana.nf
YannVRB: Fixed, same way as paired_savana.nf.

subworkflows/local/paired/paired_savana.nf
YannVRB: Fixed, now joins on SAVANA_CLASSIFY.out.somatic_vcf with remainder:true, falling back to [] when a sample has no somatic SVs (it's optional:true) rather than dropping the sample from CNA entirely.

subworkflows/local/paired/paired_savana.nf
YannVRB: Dropped.

subworkflows/local/paired/paired_savana.nf
YannVRB: Fixed, SAVANA's input is now built from PHASING_HAPLOTYPING.out.tumor_normal_hapbams_ch instead of severus_input's unphased BAMs.

subworkflows/local/tumor_only/tumoronly_savana.nf
YannVRB: Fixed, tumor-only now passes [] for snp_vcf and a genome-specific --g1000_vcf (1000g_hg38/1000g_t2t) instead. Confirmed on a real CHM13 run: sample3 gets --g1000_vcf 1000g_t2t, no --snp_vcf. Left the gnomAD-SV/HMF PoN filtering suggestion as a follow-up, as you flagged.

workflows/lrsomatic.nf
YannVRB: Fixed, prefixed with _.

workflows/lrsomatic.nf
YannVRB: savana_somatic_vcf is now wired into a new VEP_SAVANA annotation call (mirrors how Severus's SV VCF already feeds SV_VEP). Dropped savana_cn_calls entirely rather than naming it, CNA output is auto-published via conf/modules.config same as ASCAT's, which doesn't get a named workflow-level variable either.

workflows/lrsomatic.nf
YannVRB: Trimmed to what runs and the tumor-only/paired split.

conf/modules.config
YannVRB: Platform switch added for SAVANA_CLASSIFY/SAVANA_TO, keyed on meta.platform, plus a new savana_pb_minsupport param (default 10) for the PacBio --min_support bump you mentioned. --contigs added too, had to patch savana/run since it had no typed input for it (unlike to/cna); genome-specific file wired via conf/igenomes.config. Left the variants/savana vs. ASCAT directory-layout question as-is for now rather than restructuring publish paths in this PR, open to a follow-up if you'd rather split it out.

conf/test.config
YannVRB: Already replied, see the earlier draft.

conf/test.config
YannVRB: Trimmed the duplicated comment. On removing errorStrategy='ignore' once --chromosomes is "fixed", see my reply on that thread, I ended up keeping --chromosomes 19 since it wasn't the actual bug. The CNA sparsity failure is real and independent of the bugs fixed in this round (confirmed via two real runs), so errorStrategy='ignore' and the untested-tumor-only-SAVANA caveat both still stand for now.

conf/test_chm13.config
YannVRB: This is actually true now, SAVANA has a real genome-conditional path since --contigs/--g1000_vcf are genome-specific (conf/igenomes.config). Kept the full process-block duplication rather than switching to includeConfig 'test.config' + overrides, since test_full.config uses the same duplication pattern, happy to restructure both together if you'd rather.

conf/test_chm13.config
YannVRB: Noted directly in the config header now. Still blocked on the test-datasets PR merging first.

tests/chm13.nf.test
YannVRB: Fixed along with the underlying bug, confirmed passing on a real -profile test_chm13 run.

tests/chm13.nf.test
YannVRB: Trimmed to one line, fuller reasoning moved to the PR description.

CHANGELOG.md
YannVRB: Fixed → #189.

modules.json
YannVRB: Reverted to dev's exact formatting by hand (no pre-commit/prettier available in my environment), down to the 4 real new entries plus the version-bump patch below.

modules/nf-core/savana/run/main.nf
YannVRB: Bumped run/classify to 1.3.8 to match cna/to, new patches for both, confirmed versions.yml now reports 1.3.8 across all four on a real run.

docs/output.md
YannVRB: Added sample_read_counts_<mode>_log2r_segmented.tsv and 10kbp_bin_ref_all_sample.bed to the tree and table.

docs/usage.md
YannVRB: Documented the platform switch, contigs, and g1000_vcf behavior, plus the new savana_pb_minsupport param.

tests/chm13.nf.test
```diff
@@ -0,0 +1,38 @@
+nextflow_pipeline {
+
+    name "Test pipeline - CHM13 reference genome"
+    script "../main.nf"
+    tag "pipeline"
+    tag "extended"
+
+    test("-profile test_chm13") {
+
+        options "-profile test_chm13"
```
ljwharbers: **Blocking: this test cannot run under nf-test.** `nf-test.config` sets `profile = "test"` and CI adds `--profile=+docker`; nf-test's `NextflowCommand.java:193-206` emits `-profile test,docker` and then appends `options`, so the command carries a second `-profile test_chm13`. Nextflow refuses that. Reproduced locally on this branch:

```
$ nextflow config -profile test -profile test_chm13 .
Can only specify option -profile once.
```

Fix: drop `options` and set `input`, `fasta` and `genome = 'CHM13'` in the `params {}` block, on top of the `test` profile. `conf/test_chm13.config` then becomes unnecessary (or shrinks to `includeConfig 'test.config'` plus those three params), which also settles the round-1 duplication comment. Since this is tagged `extended` it only runs on pushes to `dev`, where it would currently fail.

tests/chm13.nf.test
```diff
@@ -0,0 +1,38 @@
...
+
+        when {
+            params {
+                outdir = "$outputDir"
+            }
+        }
+
+        then {
+            assertAll(
+                { assert workflow.success},
```
ljwharbers, line 21: `nf-core pipelines lint` fails `nf_test_content` here: "'tests/chm13.nf.test' does not snapshot a 'versions.yml' file". Mirror the `then` block of `tests/default.nf.test` (versions snapshot with `removeNextflowVersion`, plus the `.nftignore`-filtered path list).

conf/test.config
```diff
@@ -40,6 +40,17 @@ process {
             "--regions 'chr19'"
        }
     }
+    // SAVANA's cellularity-fitting step needs more real het-SNPs than this minimal chr19 slice
+    // has; nf-core's own module test suite marks the same output optional for that reason.
+    withName: '.*SAVANA_CNA' {
```
ljwharbers, line 50: **Blocking: CNA never completes in tests, and the cause is tunable.** The regenerated `tests/default.nf.test.snap` has no CNA outputs and no `SAVANA_CNA` in the versions block, so `savana cna` fails for both paired samples in every run and `errorStrategy = 'ignore'` hides it. Same for `SAVANA_TO` (no `sample3/variants/savana`), which means `--g1000_vcf`, `--contigs`/`--chromosomes` on `to`, and all CNA wiring are unverified anywhere.

The traceback you quoted is not het-SNP sparsity but a coverage floor: `allele_counter.run_interval` (`allele_counter.py:208-250`) only emits positions with `DP >= allele_min_reads` (default 10, `savana.py:519/613`). On the downsampled chr19 BAMs no window passes, so no temp files are written, `concatenate_sort_temp_files_and_write` prints "No temporary files found" without creating the INTERIM bed, and `process_allele_counts` raises exactly that `FileNotFoundError`.

Ask: add `--allele_min_reads 2` (and `--allele_mapq 0` if still needed) to the test-profile `ext.args` for `SAVANA_CNA` and `SAVANA_TO`, re-run both profiles, and if they complete (a failed fit is handled gracefully via `No_fit_found_PARAMS.tsv`) drop `errorStrategy = 'ignore'` and assert `sample3/variants/savana/sample3.classified.somatic.vcf`. If CNA genuinely cannot complete on this data, at least keep tumor-only SV calling testable: `savana to` skips CNA when no allele source is given (`savana.py:371`), so make the g1000 choice overridable in the test profile instead of hard-setting it from igenomes inside the workflow. As in round 1: `ignore` should survive only on `SAVANA_CNA`, and only for a genuine fitting error.

conf/test.config
```diff
@@ -40,6 +40,17 @@ process {
...
+       ext.args = {
+            "--chromosomes 19"
+       }
+       errorStrategy = 'ignore'
+    }
+    withName: '.*SAVANA_TO' {
```
ljwharbers, line 53: Once the CNA floor above is addressed this should go, together with the matching block in `test_chm13.config`. With it in place tumor-only SAVANA is untested end to end (the test explicitly does not assert on sample3).

Also: `savana_chromosomes` is set in `params` below and reaches `SAVANA_TO` via `modules.config`, but `SAVANA_CNA` still gets a hardcoded `--chromosomes 19` here and in `test_chm13.config`. Finish the mechanism: give `.*:SAVANA_CNA` the same `ext.args` closure reading the param in `modules.config`, and let the test configs only set the param.

conf/igenomes.config
```diff
@@ -39,6 +41,8 @@ params {
             bed_file        = "https://raw.githubusercontent.com/KolmogorovLab/Severus/refs/heads/main/vntrs/chm13.bed"
             vep_genome      = "T2T-CHM13v2.0"
             vep_species     = "homo_sapiens_gca009914755v4"
+            savana_contigs  = "https://raw.githubusercontent.com/IntGenomicsLab/test-datasets/main/references/savana/contigs.chr.chm13.txt"
```
ljwharbers, line 44: Now resolves (test-datasets #3 merged), so no longer a 404, but production reference config should not point at a test-data repository, and the hg38 line above tracks SAVANA's unpinned `main` branch. Both files are 24 lines. Please commit them as `assets/savana/contigs.chr.{hg38,chm13}.txt` and reference `${projectDir}/assets/...` here.

workflows/lrsomatic.nf
```diff
@@ -941,6 +946,127 @@ workflow LRSOMATIC {
...
+
+        savana_tumoronly_hapbams
+            .mix(savana_paired_hapbams)
+            .join(PHASING_HAPLOTYPING.out.phased_germline_vcf)
+            .set { savana_input }
+        // savana_input: [meta, tumor_bam, tumor_bai, normal_bam, normal_bai, phased_germline_vcf, phased_germline_tbi]
+        //   normal_bam/bai are [] for tumor-only samples
+
+        savana_input
+            .branch { meta, _tumor_bam, _tumor_bai, normal_bam, _normal_bai, _phased_vcf, _phased_tbi ->
```
ljwharbers, line 1003: New `nextflow lint` warning, the only one this PR adds relative to `dev` (15 -> 16): `Parameter was not used -- prefix with _`. Rename `meta` to `_meta`.

Related simplification: `savana_input` is branched twice, first on `meta.paired_data` (the hapbams branch above) and again here on `normal_bam`. Build `tumoronly_savana_input` and the paired input directly from the first branch and drop this second `.branch`.

conf/modules.config
```diff
@@ -431,6 +431,29 @@ process {
...
+        publishDir = [
+            path: { "${params.outdir}/${meta.id}/variants/savana" },
+            mode: params.publish_dir_mode,
+            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
+        ]
+    }
+    withName: '.*:SAVANA_CLASSIFY' {
```
ljwharbers, line 441: `SAVANA_RUN` has no PacBio switch. `savana run` has no `--pb` (`savana.py:424-439`), so it always calls with `min_support = 3` unless told otherwise, whereas SAVANA's combined `to`/main flow applies the PacBio `min_support` at the run stage too (`savana.py:321-328`). Classify filters the final set at 10, so the effect is confined to the raw `sv_breakpoints.vcf` FILTER column and runtime, but for consistency add

```groovy
withName: '.*:SAVANA_RUN' {
    ext.args = { meta.platform == 'pb' ? "--min_support ${params.savana_pb_minsupport}" : '' }
}
```

conf/modules.config
```diff
@@ -613,5 +636,14 @@ process {
             saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
         ]
     }
+    withName : '.*:VEP_SAVANA' {
```
ljwharbers, line 653: Naming: the existing aliases are `SOMATIC_VEP`, `GERMLINE_VEP`, `SV_VEP` with prefixes `<id>_SOMATIC_VEP` etc. `VEP_SAVANA` / `<id>_VEP_SAVANA` inverts that; `SAVANA_VEP` / `<id>_SAVANA_VEP` would match (also the include alias in `workflows/lrsomatic.nf:36`).

nextflow.config
```diff
@@ -71,6 +72,10 @@ params {
     // Severus options
     severus_minsupport          = 3
 
+    // Savana options
+    savana_pb_minsupport        = 10
+    savana_chromosomes          = null // e.g. '19' to restrict SAVANA_TO to one chromosome (test profiles only)
```
ljwharbers: A parameter exposed in `nextflow_schema.json` should not be documented as "(test profiles only)". Once `SAVANA_CNA` reads it too (see `conf/test.config`), describe it as what it is: a restriction of SAVANA's allele counting and binning to given chromosome indices, and update the schema text at `nextflow_schema.json:244`, which currently mentions only `savana to`.

modules/nf-core/savana/run/main.nf
```diff
@@ -0,0 +1,53 @@
...
+
+    conda "${moduleDir}/environment.yml"
+    container "${workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/savana:1.3.8--pyhdfd78af_0'
+        : 'quay.io/biocontainers/savana:1.3.8--pyhdfd78af_0'}"
+
+    input:
+    tuple val(meta), path(tumour), path(tumour_index), path(normal), path(normal_index)
+    tuple val(meta2), path(ref), path(ref_index)
+    tuple val(meta3), path(contigs)
```
ljwharbers, line 13: The patch adds this input but `meta.yml` was not updated, and `run/tests/main.nf.test.snap` and `classify/tests/main.nf.test.snap` still record `savana: 1.3.7`. `nf-test.config` ignores module tests so CI will not notice, but the module is now internally inconsistent. Update `meta.yml`, regenerate the snapshots (or say why not), and re-run `nf-core modules patch savana/run` so the `.diff` covers it.

subworkflows/local/paired/paired_savana.nf
```diff
@@ -0,0 +1,81 @@
...
+        .map { meta, tumor_bam, tumor_bai, normal_bam, normal_bai, phased_vcf, allele_counts_het_snps, somatic_vcf ->
+            return [meta, tumor_bam, tumor_bai, normal_bam, normal_bai, phased_vcf, allele_counts_het_snps, somatic_vcf ?: []]
+        }
+        .set { savana_cna_input }
+    // savana_cna_input: [meta, tumor_bam, tumor_bai, normal_bam, normal_bai, snp_vcf, [], breakpoints]
+
+    SAVANA_CNA (
+        savana_cna_input,
+        ch_fasta_fai,
+        [],  // contigs -- SAVANA_CNA's own module doesn't take a genome-aware contigs file; test profile scopes it via ext.args instead
```
ljwharbers, line 73: Misleading: the module does take a `contigs` path, but `savana cna` has no `--contigs` flag at all (`cna_parser`, `savana.py:505-524`), so wiring it would error. Either patch `--contigs` out of the `cna` module or reword to say that.

docs/output.md
```diff
@@ -339,6 +339,51 @@ Present in **tumor-only** samples (no matched normal).
...
+| `sample.classified.germline.vcf`               | Classified germline SVs                                                          |
+| `sample_segmented_absolute_copy_number.tsv`    | Segmented absolute copy-number calls from `savana cna`/`savana to`               |
+| `sample_ranked_solutions.tsv`                  | Candidate purity/ploidy solutions, ranked                                        |
+| `sample_fitted_purity_ploidy.tsv`              | Selected purity/ploidy fit                                                       |
+| `sample_raw_read_counts.tsv`                   | Raw binned read counts used for CN segmentation                                  |
+| `sample_read_counts_<mode>_log2r_segmented.tsv`| Segmented log2 read-count ratios per bin (`<mode>` is `mnorm` or `self`)          |
+| `sample_allele_counts_hetSNPs.bed`             | Heterozygous-SNP allele counts (only when SNP/allele-frequency input is given)    |
+| `10kbp_bin_ref_all_sample.bed`                 | Genome bins used for read-count binning                                          |
```
ljwharbers: In tumor-only mode the file is `10kbp_bin_ref_all_<sample>_with_SV_breakpoints.bed` (see the `to` module stub, `modules/nf-core/savana/to/main.nf:85`); tree and table only show the paired name.

ljwharbers (commented):
## Round 2 review (b3b0b27)

Thanks for the thorough follow-up. I re-checked the final head against the SAVANA 1.3.8 source, nf-test's source, and local `nextflow lint`, `pre-commit` and `nf-core pipelines lint` runs.

**Retraction first.** My round-1 comment on `--chromosomes 19` was wrong and your reading is right: `-c/--chromosomes` is a 1-based index into a hardcoded 24-entry list (`bin_generator.py:137-142`, `allele_counter.py:181-186`), chr-prefixed when the FASTA uses `chr`, so `19` -> `chr19`; the default `all` would `KeyError` on a chr19-only reference. Keep it.

**Verified fixed**
- One-sample bug (`.first()`), CNA now fed the classified somatic VCF (matches `savana.py:369/409`), haplotagged BAMs with meta handling mirroring `severus_input`.
- `--pb`/`--ont` on classify/to. sample2 is PacBio, so the `--pb --min_support 10` branch actually ran in CI.
- `--g1000_vcf` for tumor-only. The 1000G VCFs are in the biocontainer (bioconda builds from the GitHub tarball; `MANIFEST.in` grafts `savana/`, including `1K_genome_vcf/`).
- `--contigs` on run/to, 1.3.8 everywhere, `modules.json` SHAs are real nf-core commits, whole-file reformat reverted, CHANGELOG/CITATIONS/docs.
- test-datasets #3 has merged, so the CHM13 fixture 404s are gone.

**Still blocking**
1. `tests/chm13.nf.test` cannot run under nf-test at all: nf-test already emits `-profile test,<engine>` and Nextflow rejects a second `-profile` (details inline). Fixing this also removes the need for `conf/test_chm13.config`.
2. CI is red on both jobs at this commit. Lint: prettier reformats `docs/output.md` and `docs/usage.md` (`pre-commit run --all-files` -> "files were modified by this hook"); `nf-core pipelines lint` additionally fails `nf_test_content` for `tests/chm13.nf.test` (no `versions.yml` snapshot). nf-test: the logs have expired and I could not recover the cause, so it needs a green re-run after the fixes.
3. The CNA half and the whole tumor-only path have never completed in either profile (no CNA files, no `sample3/variants/savana`, no `SAVANA_CNA`/`SAVANA_TO` in the snapshot's versions block), hidden by `errorStrategy = 'ignore'`. The cause is a tunable coverage floor, not het-SNP sparsity; details and a concrete ask inline on `conf/test.config`.

**Should fix** (inline): new `nextflow lint` warning at `workflows/lrsomatic.nf:1003`; `savana_chromosomes` reaches `SAVANA_TO` but `SAVANA_CNA` still hardcodes `--chromosomes 19`; `SAVANA_RUN` has no PacBio switch; `savana/run` `meta.yml` and the run/classify module snapshots were not updated with the patch; production reference config points at test-datasets and SAVANA's `main` branch for the contigs files; PR description/checklist still empty.

**Minor** (inline): misleading `contigs` comment in `paired_savana.nf`; `VEP_SAVANA` naming; double branching of `savana_input`; tumor-only bin filename in `docs/output.md`; README author line; the `variants/savana` vs ASCAT layout question is still open, decide or explicitly defer in the PR body.

conf/test.config
ljwharbers: Retracting this comment: you are right. `-c/--chromosomes` indexes a hardcoded 24-entry chromosome list (`bin_generator.py:137-142`, `allele_counter.py:181-186`), not the reference contigs, so `19` resolves to `chr19` and the default `all` would `KeyError` on the chr19-only test reference. Keep `--chromosomes 19`. The actual CNA failure is the `allele_min_reads` coverage floor, see my round-2 comment on this file.

ljwharbers:
Summary of the round-2 review (details are in the inline comments on b3b0b27).

**Round 1 is resolved**, and my `--chromosomes 19` comment was wrong: it is a 1-based index into SAVANA's hardcoded 24-chromosome list, so `19` is `chr19` and the default `all` would crash on the chr19-only reference. Keep it. Verified as fixed: the one-sample `combine` bug, CNA fed the classified somatic VCF, haplotagged BAMs, `--pb`/`--ont` on classify/to (sample2 is PacBio so that branch ran in CI), `--g1000_vcf` for tumor-only (the 1000G VCFs are shipped in the biocontainer), `--contigs` on run/to, 1.3.8 everywhere, `modules.json` cleaned up, docs/CHANGELOG/CITATIONS. test-datasets #3 has merged, so the CHM13 fixture 404s are gone.

**Blocking**
1. `tests/chm13.nf.test` cannot run under nf-test. nf-test already passes `-profile test,<engine>` and appends the test's `options "-profile test_chm13"`; Nextflow refuses a second `-profile` (`Can only specify option -profile once`, reproduced locally). Set `input`, `fasta`, `genome = 'CHM13'` in the test's `params {}` instead. `conf/test_chm13.config` then becomes unnecessary.
2. CI is red on both jobs at this commit. Lint: prettier reformats `docs/output.md` and `docs/usage.md` (`pre-commit run --all-files`), and `nf-core pipelines lint` fails `nf_test_content` for `tests/chm13.nf.test` (no `versions.yml` snapshot). nf-test: logs expired, cause unknown, needs a green re-run.
3. CNA and the whole tumor-only path never complete in either profile (no CNA files, no `sample3/variants/savana`, no `SAVANA_CNA`/`SAVANA_TO` in the snapshot versions), hidden by `errorStrategy = 'ignore'`. The cause is SAVANA's `allele_min_reads` coverage floor (default 10) on the downsampled BAMs, which yields no INTERIM bed and the `FileNotFoundError` you quoted. Please try `--allele_min_reads 2` in the test-profile `ext.args` for `SAVANA_CNA`/`SAVANA_TO`; if they complete, drop the `ignore` and assert on sample3's somatic VCF.

**Should fix**
- New `nextflow lint` warning at `workflows/lrsomatic.nf:1003` (unused `meta`).
- `savana_chromosomes` reaches `SAVANA_TO` only; `SAVANA_CNA` still hardcodes `--chromosomes 19` in both test configs.
- `SAVANA_RUN` has no PacBio switch (`savana run` has no `--pb`; pass `--min_support ${params.savana_pb_minsupport}` for `pb`).
- `savana/run` patch: `meta.yml` not updated, run/classify snapshots still say 1.3.7.
- Contigs files: commit them to `assets/savana/` instead of pointing production igenomes at test-datasets and SAVANA's unpinned `main`.
- PR description and checklist are still empty.

**Minor**: misleading `contigs` comment in `paired_savana.nf` (`savana cna` has no `--contigs`), `VEP_SAVANA` naming vs `SV_VEP` etc., redundant second `.branch` on `savana_input`, tumor-only bin filename in `docs/output.md`, README author line, and the `variants/savana` vs ASCAT layout question is still open.

ljwharbers:
Looks good to me, thanks a lot for implementing this! You're welcome to merge whenever you want.

ljwharbers (approved):

Issue #192, opened 2026-09-09
Add SAVANA, padfoot and ReConPlot


Tim-Yu:
### Description of feature

Hi,

I am currently working on adding SAVANA, padfoot and ReConPlot modules to the workflow as I found that the CN calling from ascat is not quite aligning to the SV reported as well as to the paired short reads. 

Currently, I am testing on our cluster and will update with the progress if you think those modules could be beneficial.

Best,

Boyu

ljwharbers:
Hi Boyu!

That would be amazing. We currently have a PR open to implement SAVANA https://github.com/IntGenomicsLab/lrsomatic/pull/189, so it would be great if you could have a look at the current implementation there and see if you would require any changes there. In terms of padfoot and ReConPlot I think these would be nice additions. What do you think? @robert-a-forsyth 

Best,
Luuk

Tim-Yu:
Thank you. Will see how #189 cope with the other modules. I just found there is a nf-core savana. I was making a local version.

ljwharbers:
Hi Boyu,

Savana is now implemented. Padfoot and ReConPlot are not yet on our to-do but I can put it on there. you are of course more than welcome to help with the implementation and open a PR for it as well :D 

Best,
Luuk

Tim-Yu:
Hi Luuk,

Just saw the PR. I will then modify my version for this Savana module then. Will PR later for the Padfoot and ReConPlot.

Best wishes,

Boyu

Tim-Yu:
Hi Luuk,

I took some time to rebase my Padfoot and ReConPlot work on top of the savana PR #189 and opened it as a draft PR #202 so you can look before I mark it ready.

Padfoot annotates somatic SVs + CNAs and ReConPlot draws the figures, each once per caller pair that produced output for a sample (Severus + Wakhan; ASCAT + Severus; SAVANA SV and CNV on its own), paired and tumour-only, on by default with `--skip_padfoot` / `--skip_reconplot`. 

[...]

PR #193, opened 2026-09-09
Vep plugins

AmberVerhasselt:
## PR checklist

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/IntGenomicsLab/lrsomatic/tree/main/docs/CONTRIBUTING.md)
- [ ] Make sure your code lints (`nf-core pipelines lint`).
- [ ] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).


workflows/lrsomatic.nf
```diff
@@ -107,6 +109,13 @@ workflow LRSOMATIC {
     params.sigprofiler_genome = getGenomeAttribute('sigprofiler_genome')
     params.sigprofiler_genome_url = getGenomeAttribute('sigprofiler_genome_url')
 
+    // vep_plugin_args is read back by conf/modules.config
+    validateVepPluginParams()
+    params.vep_plugin_args = resolveVepPlugins().args
```
ljwharbers, line 114: **Blocking: no end-to-end run.** Every commit message verifies with `-preview`, which never runs a task, and 5af0129 confirms the protein table was only exercised via `tabix`, not through VEP. Before merge please run one GRCh38 and one CHM13 sample through and paste the counts, e.g.

```
bcftools +split-vep -f '%CADD_PHRED\t%REVEL\t%am_pathogenicity\t%ClinVar_CLNSIG\n' sample_SOMATIC_VEP.vcf.gz | grep -vc '^\.'
```

for GRCh38, and `%SIFT_score %PolyPhen_humvar_score %am_pathogenicity %AlphaMissenseProtein_match` for CHM13. The distribution of `AlphaMissenseProtein_match` is the real test of the protein-space lookup.

Also add nf-test function tests for `resolveVepPlugins()` / `validateVepPluginParams()` (pattern: `subworkflows/nf-core/utils_nfcore_pipeline/tests/main.function.nf.test`) so precedence and the error paths stay covered. CI sets `skip_vep_plugins = true`, so none of this is tested today.

conf/modules.config
```diff
@@ -596,6 +596,13 @@ process {
         ]
     }
 
+    withName : '.*:WGET_(REVEL|EVE)' {
```
ljwharbers, line 599: **Blocking: prepared files are published.** `VEPPLUGIN_REVEL` / `VEPPLUGIN_EVE` have no `withName` block, so the top-level default `publishDir` applies and `revel_grch38.tsv.gz{,.tbi}` lands in `${outdir}/vepplugin/`. docs/usage.md ("Prepared files are **not** published to `--outdir`") and the CHANGELOG say otherwise.

Either add

```groovy
withName: '.*:VEPPLUGIN_.*' {
    publishDir = [ enabled: false ]
}
```

or, my preference and matching `SIGPROFILER_INSTALL`, publish deliberately to `${outdir}/vep_plugins/` and document it as the way to reuse the file via `--vep_revel` / `--vep_revel_tbi` on later runs.

Separately: `WGET_EVE` inherits `process_single` (4 h). The docs say the 9.6 GB EVE archive serves at a few hundred KB/s, i.e. many hours, so attempt 1 will time out. Give it `process_long` or a `time` override.

workflows/lrsomatic.nf
```diff
@@ -107,6 +109,13 @@ workflow LRSOMATIC {
...
+
+    vep_custom = params.vep_custom != null ? file(params.vep_custom) : []
```
ljwharbers, line 116: **Blocking: `--vep_custom` can clobber ClinVar.** The module rewrites the first `--custom file=` in `ext.args` to the staged `custom_vep`. `ext.args` is now `vep_args + vep_plugin_args`, so a user passing `--vep_custom X` without a `--custom file=` entry of their own in `--vep_args` (the docs never told anyone to add one) gets the *ClinVar* entry rewritten: ClinVar is silently replaced by their VCF while keeping `short_name=ClinVar,fields=CLNSIG...`. Before this PR that combination was a no-op.

Suggest a check in `validateVepPluginParams()`:

```groovy
if (params.vep_custom && !(params.vep_args =~ /--custom file=/)) {
    error("--vep_custom: needs a matching '--custom file=...' entry in --vep_args")
}
```

or give the ClinVar entry a form the rewrite cannot match.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - [#176](https://github.com/IntGenomicsLab/lrsomatic/pull/176) - Vendored the [lrsomatic_report](https://github.com/ljwharbers/lrsomatic_report) v1.3.2 tool source at `assets/lrsomatic_report`, so `nextflow run IntGenomicsLab/lrsomatic` ships it without a submodule checkout (@ljwharbers).
 - [#176](https://github.com/IntGenomicsLab/lrsomatic/pull/176) - Added a `solution_dirs` output to the WAKHAN module so its per-solution copy-number plots can be staged downstream (@ljwharbers).
+- [#192](https://github.com/IntGenomicsLab/lrsomatic/pull/192) - Added VEP plugins for AlphaMissense, SIFT/PolyPhen, ClinVar, CADD and REVEL on GRCh38, and AlphaMissense plus SIFT/PolyPhen on CHM13 via protein-space lookup. Enabled by default with `--genome GRCh38` or `--genome CHM13`: the pipeline fetches each resource from its original source and prepares the releases VEP cannot read as published, so a first GRCh38 run downloads around 83 GB. Any `--vep_*` path overrides its default, and `--skip_vep_plugins` turns the set off. EVE is wired up but left opt-in behind `--vep_eve`. Nothing is redistributed or published to the output directory (@AmberVerhasselt).
```
ljwharbers, line 14: **Blocking (trivial):** this PR is #193, not #192 (#192 does not exist).

Also "Nothing is redistributed" is stale since 5af0129: the AlphaMissense GRCh38 `.tbi` and the CHM13 protein table are hosted on the lab Globus endpoint. The same sentence is in docs/usage.md under "VEP plugins" ("The pipeline hosts and redistributes none of this data").

conf/igenomes.config
```diff
@@ -22,6 +22,15 @@ params.genomes = [
         bed_file        : "https://raw.githubusercontent.com/KolmogorovLab/Severus/refs/heads/main/vntrs/human_GRCh38_no_alt_analysis_set.trf.bed",
         vep_genome      : "GRCh38",
         vep_species     : "homo_sapiens",
+        vep_alphamissense        : "https://storage.googleapis.com/dm_alphamissense/AlphaMissense_hg38.tsv.gz",
+        vep_alphamissense_tbi    : "https://g-608c0c.273595.03c0.data.globus.org/VEP_plugins/AlphaMissense_hg38.tsv.gz.tbi",
+        vep_clinvar              : "https://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh38/clinvar.vcf.gz",
```
ljwharbers, line 27: **Should fix: unpinned ClinVar.** `clinvar.vcf.gz` is the rolling weekly file, and its `.tbi` is a separate download, so two runs a week apart annotate differently and a run that straddles an NCBI update can pair a new VCF with an old index. Please pin a dated file from `vcf_GRCh38/archive_2.0/<year>/`, as the CHM13 entry already does with `clinvar_20240624`.

Same on line 53: `current_variation/` moves at every Ensembl release; pin `release-115/variation/pangenomes/...`.

modules/local/vepplugin/revel/main.nf
```diff
@@ -0,0 +1,34 @@
...
+    tuple val("${task.process}"), val('tabix'), eval("tabix -h 2>&1 | grep -oP 'Version:\\s*\\K[^\\s]+'"), topic: versions, emit: versions_tabix
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    """
+    # The re-sort onto the GRCh38 column needs scratch space rather than /tmp
+    export TMPDIR=\${TMPDIR:-\$PWD}
```
ljwharbers, line 24: **Should fix:** `export TMPDIR=${TMPDIR:-$PWD}` keeps the host's `TMPDIR` when apptainer forwards the environment, and on Mindwell that path is not bound inside the container, so `mktemp -d` fails. This is the same failure we hit with Clair3 under nf-test. The task dir is already on scratch, so just `export TMPDIR=$PWD` (or pass `-T .` to sort). Same in `modules/local/vepplugin/eve/main.nf:24`.

Nit: the container is `bcftools_htslib` while environment.yml lists only htslib; pick one.

conf/igenomes.config
```diff
@@ -22,6 +22,15 @@ params.genomes = [
...
+        vep_clinvar_tbi          : "https://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh38/clinvar.vcf.gz.tbi",
+        vep_cadd_snv             : "https://krishna.gs.washington.edu/download/CADD/v1.7/GRCh38/whole_genome_SNVs.tsv.gz",
```
ljwharbers, line 29: **Design question for @ljwharbers: CADD on by default.** 81 GB CADD as a default means every non-`-resume` session re-stages ~83 GB into `work/stage-<session-id>/` (Nextflow's foreign-file cache is per session, not per work dir), through the head job, with no resumable download. On a 500 GB scratch quota that is a lot per run.

Options: (a) make CADD opt-in like EVE; (b) keep it on but ship a lab params file pointing `--vep_cadd_snv*` at a one-time copy on project storage. I'd lean (a) for the public default.

assets/vep_plugins/AlphaMissenseProtein.pm
```diff
@@ -0,0 +1,260 @@
...
+  my $headers = $self->{headers};
+  $data{$headers->[$_]} = $split[$_] for 0 .. $#$headers;
+
+  return \%data;
+}
+
+sub get_start { return $_[1]->{aapos}; }
+sub get_end   { return $_[1]->{aapos}; }
+
+sub run {
```
ljwharbers, line 204: **Should fix:** `aa_mismatch` is meant to flag CHM13/GRCh38 protein divergence, but stop-gain, start-lost and stop-lost single-residue changes (`A/*`, `M/L` at position 1) also fall through to `aa_mismatch`, because the table cannot contain them. Restrict the way REVEL.pm does:

```perl
return {} unless grep { $_->SO_term eq 'missense_variant' } @{$tva->get_all_OverlapConsequences};
```

docs/usage.md
```diff
@@ -372,6 +402,169 @@ To further assist in reproducibility, you can use share and reuse [parameter fil
...
+  publishes licence terms that clearly permit distributing a derived table. AlphaMissense covers the
+  same class of variant and is CC BY 4.0, so it is used instead.
+- **SpliceAI** — not currently wired up on either assembly.
+
+> [!IMPORTANT]
+> **CADD and REVEL are enabled by default and are free for non-commercial use only**; EVE, if you
+> enable it, is the same. The pipeline cannot accept those terms on your behalf: if your work is
+> commercial, pass `--skip_vep_plugins`, or set only the resources you are licensed for.
+
+> [!IMPORTANT]
```
ljwharbers, line 562: **Nit:** VEP does map `1` <-> `chr1` for `--custom` files (`BaseVEP::get_source_chr_name` tries synonyms, then adds/strips `chr`), so "a mismatch produces empty annotation" overstates it. Keep the advice, soften the claim.

Also under "Downloading and reusing the data": it is `-resume` (same session id) that avoids the re-download, not the work directory. A fresh session in the same work dir downloads again because the stage dir is `work/stage-<session-id>/`.

Checklist items still open: `docs/output.md` should list the new CSQ fields (`am_pathogenicity`, `am_class`, `CADD_PHRED/RAW`, `REVEL`, `ClinVar_CLNSIG/CLNREVSTAT/CLNDN`, `SIFT_*`, `PolyPhen_*`, `AlphaMissenseProtein_match`) and when they appear, and README's tool list should name the new tools. The PR description is still the unfilled template.

subworkflows/local/prepare_vep_plugins.nf
```diff
@@ -0,0 +1,98 @@
...
+include { VEPPLUGIN_EVE        } from '../../modules/local/vepplugin/eve/main.nf'
+
+include { resolveVepPlugins    } from './utils_nfcore_lrsomatic_pipeline'
+
+workflow PREPARE_VEP_PLUGINS {
+
+    main:
+
+    // Resolved here rather than taken as an input: a workflow input would arrive in a channel
+    def plugins = resolveVepPlugins()
```
ljwharbers, line 19: **Nit:** `resolveVepPlugins()` runs here and again at lrsomatic.nf:114, so every launch does two rounds of `checkIfExists` HEAD requests on ~10 URLs. The resolved map could be passed in; a plain Groovy value works as a workflow input.

bin/prepare_vep_plugin_data.sh
```diff
@@ -0,0 +1,120 @@
...
+    local src=${1:-}
+    local outdir=${2:-.}
+    [[ -n $src ]] || die "usage: $0 revel <revel-v1.3_all_chromosomes.zip|unpacked-dir> [outdir]"
+    [[ -r $src ]] || die "cannot read $src"
+    need bgzip tabix awk sort
+
+    mkdir -p "$outdir"
+    local work
+    work=$(mktemp -d "${TMPDIR:-/tmp}/revel.XXXXXX")
+    trap 'rm -rf "$work"' RETURN
```
ljwharbers, line 38: **Nit:** a `RETURN` trap does not fire when `set -e` exits the script mid-function, so the temp dir leaks on failure. `trap ... EXIT` covers both.

subworkflows/local/utils_nfcore_lrsomatic_pipeline/main.nf
```diff
@@ -345,6 +345,195 @@ def getGenomeAttribute(attribute) {
...
+        'vep_revel'           : 'vep_revel_tbi',
+        'vep_eve'             : 'vep_eve_tbi'
+    ]
+}
+
+//
+// The index for a plugin data file, or null when there is none. Overriding the data file drops the
+// default index, which was built from different content
+//
+def vepPluginIndex(data_param) {
```
ljwharbers, line 393: **Nit:** a user-supplied `--vep_revel_tbi` / `--vep_eve_tbi` is silently ignored when the data param is a `.zip`. Worth an error, since the user clearly expected it to be used.

subworkflows/local/utils_nfcore_lrsomatic_pipeline/main.nf
```diff
@@ -345,6 +345,195 @@ def getGenomeAttribute(attribute) {
...
+    return [
+        'vep_revel': 'revel_grch38.tsv.gz',
+        'vep_eve'  : 'eve_merged.vcf.gz'
+    ][data_param]
+}
+
+//
+// Exit if the plugin params contradict each other or the target assembly, before any download
+//
+def validateVepPluginParams() {
```
ljwharbers, line 424: **Nit:** the non-CHM13 branch at the end of this function prints "On null use --vep_alphamissense" when `vep_genome` is unset (custom reference without `--genome`).

ljwharbers (changes requested):
Thanks, this is a big and carefully documented piece of work, and most of the plumbing checks out. I verified against the VEP 115.2 container that `--dir_plugins .` does not hide the stock plugins (`vep` does `use lib $RealBin`, so the share dir with AlphaMissense.pm, CADD.pm, REVEL.pm, EVE.pm and PolyPhen_SIFT.pm is always on `@INC`), that `DBD::SQLite` is present for PolyPhen_SIFT, that the `file=`/`snv=`/`indels=`/`db=` argument forms match the plugin sources, that the REVEL prep output has the 9-column `#`-prefixed header REVEL.pm expects, and that VEP maps `1` <-> `chr1` for both tabix plugins and `--custom` files.

Four things I'd like resolved before merging:

1. The plugin path has never been executed end to end, only `-preview`ed, and the SIGPIPE fix in d92b536 shows why that matters.
2. The prepared REVEL/EVE files *are* published to `--outdir` through the default `publishDir`, contradicting the docs and CHANGELOG.
3. `--vep_custom` can now silently replace the ClinVar track.
4. A few stale statements (PR number, "nothing redistributed").

The 81 GB CADD default is a design call for @ljwharbers, flagged inline. Details on everything below.

Review drafted with Claude Code.

subworkflows/local/utils_nfcore_lrsomatic_pipeline/main.nf
```diff
@@ -345,6 +345,195 @@ def getGenomeAttribute(attribute) {
...
+}
+
+//
+// Whether a resource still has to be reshaped before VEP can read it, true only for a release zip
+//
+def vepPluginNeedsPrep(data_param) {
+    def value = vepPluginResource(data_param)
+    return value \
+        && ['vep_revel', 'vep_eve'].contains(data_param) \
+        && value.toString().toLowerCase().endsWith('.zip')
```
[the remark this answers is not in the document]
AmberVerhasselt: This one is a false positive — Groovy does support `\` as a line-continuation
character, and the function compiled and ran fine. I have reformatted it to
trailing `&&` anyway, since that reads better and matches the rest of the file.

conf/modules.config
AmberVerhasselt: Gone with your preferred option: `VEPPLUGIN_REVEL` / `VEPPLUGIN_EVE` publish to
`${outdir}/vep_plugins/`, and `docs/usage.md` documents reusing them via
`--vep_revel` / `--vep_revel_tbi` on later runs.

One addition to the docs while doing it. Publishing is not redistribution — it
copies a file the user downloaded themselves onto their own disk — but
`${outdir}/vep_plugins/` sits inside the tree people tar up and send to
collaborators, and a 667 MB non-commercial REVEL table riding along in a shared
archive *would* be. So there is an `[!IMPORTANT]` callout telling people to
exclude the directory when sharing or archiving results. Say the word if you
would rather it went to the cache tree next to `ENSEMBLVEP_DOWNLOAD` and
`SIGPROFILER_INSTALL` instead, which would sidestep that entirely.

`WGET_EVE` now gets `time = { 24.h * task.attempt }`.

conf/igenomes.config
AmberVerhasselt: Both pinned, and both URLs checked to return 200:

- GRCh38 ClinVar → `archive_2.0/2026/clinvar_20260829.vcf.gz` (+ `.tbi`)
- CHM13 pangenome PolyPhen/SIFT → `release-115/variation/pangenomes/Human/…`

`docs/usage.md`'s source table is updated to match.

subworkflows/local/utils_nfcore_lrsomatic_pipeline/main.nf
AmberVerhasselt: Both are errors now. A `--vep_revel_tbi` / `--vep_eve_tbi` passed alongside a
release that still needs reshaping is rejected rather than silently dropped, and
the AlphaMissense message no longer prints "On null" when `vep_genome` is unset.
Both have test cases.

bin/prepare_vep_plugin_data.sh
AmberVerhasselt: Switched to `EXIT` in both `prep_revel` and `prep_eve`. The two are never called
in one invocation — the `case` dispatches one — so the second trap replacing the
first is harmless.

subworkflows/local/prepare_vep_plugins.nf
AmberVerhasselt: Done, and you were right that a plain Groovy value works as a workflow input —
the Map survives `take:` intact, so the subworkflow still reads `.prepare` and
`.ready_files` directly. Both call sites are in the same workflow scope, so it
is just passed down.

For the record, in case it comes up again: memoising inside the utils module
does *not* work. I tried `@groovy.transform.Memoized` and a `@Field static`
cache, and probed both with a counter — still two calls each time, because the
two include sites are separate script instances that share no state.

modules/local/vepplugin/revel/main.nf
AmberVerhasselt: Both modules now `export TMPDIR=$PWD` unconditionally.

On the container nit: I added `bcftools` to both `environment.yml` files so they
match the pinned `bcftools_htslib` image. Worth noting the script only uses
htslib tools (bgzip, tabix) plus coreutils, so an htslib-only image would be
tighter — I left that alone because it needs a new container URI and a rebuild.
Happy to do it if you would rather go that direction.

docs/usage.md
AmberVerhasselt: All four done:

- Contig-naming claim softened; it now says VEP reconciles `1`/`chr1` via
  `BaseVEP::get_source_chr_name` (synonyms, then adding/stripping `chr`), that
  this is not guaranteed for every contig, and that a name it cannot map
  annotates nothing rather than erroring.
- Reuse section now credits `-resume` and the session id, not the work
  directory, and explains `work/stage-<session-id>/`.
- `docs/output.md` tables every new CSQ field, its source and when it appears.
- README's tool list names the plugins.

The PR description still needs filling in — doing that separately.

subworkflows/local/utils_nfcore_lrsomatic_pipeline/main.nf
AmberVerhasselt: Both are errors now. A `--vep_revel_tbi` / `--vep_eve_tbi` passed alongside a
release that still needs reshaping is rejected rather than silently dropped, and
the AlphaMissense message no longer prints "On null" when `vep_genome` is unset.
Both have test cases.

assets/vep_plugins/AlphaMissenseProtein.pm
AmberVerhasselt: Fixed exactly as suggested — `run()` now returns early unless the consequence
set contains `missense_variant`, matching REVEL.pm. Stop-gain, start-lost and
stop-lost no longer surface as protein divergence. `docs/output.md` documents
what each `AlphaMissenseProtein_match` value means, and notes that anything
non-missense carries no `AlphaMissenseProtein_*` field at all.

conf/igenomes.config
AmberVerhasselt: Went with (a). `vep_cadd_snv`, `vep_cadd_indel` and their `_tbi` are out of the
GRCh38 defaults, so CADD behaves like EVE — wired up, opt-in, nothing fetched
unless asked for. A first GRCh38 run is about 1.4 GB now rather than 83 GB.

`docs/usage.md` explains why in the opt-in section (the per-session foreign-file
cache, no resumable download) and recommends downloading once to local storage
and pointing `--vep_cadd_snv*` at that, which is your option (b) available to
anyone who wants it without making it the public default.

workflows/lrsomatic.nf
AmberVerhasselt: The nf-test function tests are in:
`subworkflows/local/utils_nfcore_lrsomatic_pipeline/tests/main.function.nf.test`,
25 tests covering `resolveVepPlugins()` precedence, the index-dropping rule, the
release-zip split and every `validateVepPluginParams()` error path. They resolve
nothing remote — each case either leaves `params.genome` unset or uses a release
zip — so they need no network and run in CI despite `skip_vep_plugins = true`.

The GRCh38 and CHM13 sample runs and the `bcftools +split-vep` counts are still
outstanding and tracked separately; this PR should not merge before they are
done. I will post the counts, including the `AlphaMissenseProtein_match`
distribution, on this thread.

One thing the function tests turned up along the way: the documented way of
enabling EVE never worked. `vepPluginNeedsPrep()` keyed on a `.zip` suffix, and
`https://evemodel.org/api/proteins/bulk/download/` has no file extension, so the
URL was taken for an already-prepared file and validation then demanded a
`--vep_eve_tbi` that cannot exist. The check now asks whether the value already
*is* a prepared (bgzipped) file, which covers the endpoint, the zip and an
unpacked directory alike. Verified with `-preview -with-dag` that
`WGET_EVE → UNZIP_EVE → VEPPLUGIN_EVE` now wires up from the documented URL.

ljwharbers (commented):
Thanks Amber, this round addresses everything from the first review, including the two threads without a reply (the `--vep_custom` guard and the CHANGELOG/redistribution wording are both in). I pulled 2e7eb85f locally: `prettier --check` and `nextflow lint` are clean on the changed files, and the 25 new function tests all pass (93 s). The `versions_tabix` point Copilot raised is not an issue either, since both modules emit on `topic: versions`, which the workflow collects.

Two things gate the merge, and a handful of small fixes are inline:

1. **The branch conflicts with `dev`** (`CHANGELOG.md` and `CITATIONS.md`, from #181 which landed today). Because GitHub cannot build the merge commit, **no `pull_request` workflow has run on any of today's commits**: nf-test and nf-core lint last ran on 9ddc669a. Please merge `upstream/dev` once more so CI covers the new code.
2. **The end-to-end counts** for GRCh38 and CHM13, as agreed on the thread. The GRCh38 run has to be repeated on 2e7eb85f, since both prep modules exited non-zero before that fix, and CHM13 has not run yet.

The inline comments are the `small` tag on the new test file (without it the tests never run in CI), five stale `help_text` strings in the schema that did not follow the CADD opt-in and URL pinning changes, the `current_variation` link in CITATIONS.md, and a missing blank line in usage.md. All documentation-only apart from the tag. The PR description is still the template; fine to fill that in with the counts.

Review drafted with Claude Code.

subworkflows/local/utils_nfcore_lrsomatic_pipeline/tests/main.function.nf.test
```diff
@@ -0,0 +1,594 @@
+nextflow_function {
+
+    name "Test VEP plugin resolution functions"
+    script "../main.nf"
+    tag "subworkflows"
+    tag "subworkflows_local"
+    tag "utils_nfcore_lrsomatic_pipeline"
+    tag "vep_plugins"
```
ljwharbers, line 11: **Should fix: these tests never run in CI as tagged.** `.github/workflows/nf-test.yml` passes `--tag small` to nf-test on `pull_request` (see `get-shards`), and the workflow has no other trigger on `dev`. The file carries `subworkflows`, `subworkflows_local`, `utils_nfcore_lrsomatic_pipeline` and `vep_plugins`, so the shard step reports no tests for it. Add

```groovy
    tag "small"
```

They take about 90 s in total locally, so they belong in the small set.

nextflow_schema.json
```diff
@@ -214,6 +214,102 @@
...
+                    "type": "string",
+                    "description": "Comma-separated ClinVar INFO fields to carry into the annotation.",
+                    "fa_icon": "fas fa-list",
+                    "default": "CLNSIG,CLNREVSTAT,CLNDN"
+                },
+                "vep_cadd_snv": {
+                    "type": "string",
+                    "description": "Path or URL to the CADD SNV score file for the CADD plugin.",
+                    "fa_icon": "fas fa-file",
+                    "help_text": "Defaults to whole_genome_SNVs.tsv.gz from https://krishna.gs.washington.edu/download/CADD/v1.7/GRCh38/ on --genome GRCh38, with its published .tbi. At 81 GB this is by far the largest download a default run makes, and nothing is published to the output directory, so point this at a local copy once you have one. GRCh38 only: CADD scores non-coding positions and so has no protein-space form usable on CHM13. Free for non-commercial use only."
```
ljwharbers: **Should fix: stale help texts.** docs/usage.md was updated for the CADD opt-in and the URL pins, but the schema was not. Five strings still describe the old state:

- `vep_cadd_snv` (this line): "Defaults to whole_genome_SNVs.tsv.gz ... on --genome GRCh38", "the largest download a default run makes", and "nothing is published to the output directory". CADD has no default now, and the prepared REVEL/EVE files are published to `<outdir>/vep_plugins/`.
- `vep_cadd_indel` (line 282): "Defaults to gnomad.genomes.r4.0.indel.tsv.gz ... on --genome GRCh38". No default.
- `vep_polyphen_sift_db` (line 246): still `current_variation/`; igenomes.config pins `release-115`.
- `vep_clinvar` (line 252): "a mismatch produces empty annotation rather than an error", the claim softened in usage.md.
- `skip_vep_plugins` (line 509): "a first run downloads tens of GB". It is about 1.4 GB on GRCh38 and 14 GB on CHM13 now.

CITATIONS.md
```diff
@@ -114,6 +142,34 @@
...
+
+  On GRCh38 the score file comes from the bucket unchanged, with a tabix index we host because the
+  release ships without one. On CHM13 the pipeline distributes **an adaptation** of this data: the
+  protein-space release re-keyed from UniProt accession to gene symbol using the UniProt ID mapping,
+  split into reference/position/alternate amino-acid columns, deduplicated where several accessions
+  share a symbol, and tabix-indexed. Rows whose accession has no gene symbol are dropped. **No score
+  is altered.** CC BY 4.0 permits this adaptation and its redistribution with attribution and, as
+  given here, an indication of the changes made.
+
+- [Ensembl pangenome PolyPhen/SIFT database](https://ftp.ensembl.org/pub/current_variation/pangenomes/Human/)
```
ljwharbers: **Nit:** this still links `current_variation/`, which moves every Ensembl release; igenomes.config and the usage table now point at `release-115/`. Worth matching so the citation names the version that was actually used.

docs/usage.md
```diff
@@ -125,6 +125,9 @@ nextflow run IntGenomicsLab/lrsomatic \
 
 If you want to run with a CHM13 reference without using `--genome CHM13` (for example, via a custom FASTA or configuration), you must also specify `--vep_genome T2T-CHM13v2.0` and `--vep_species homo_sapiens_gca009914755v4`.
 
+This cache carries gene and transcript models only, so every pathogenicity score on CHM13 has to come from a plugin. With `--genome CHM13` the pipeline fetches those itself; note that only the two predictors keyed on protein rather than genome coordinates can reach the assembly at all. See [VEP plugins](#vep-plugins) for which those are and why.
+
+Setting `--vep_genome T2T-CHM13v2.0` by hand, without `--genome CHM13`, resolves no plugin defaults — supply the `--vep_*` paths yourself, or accept VEP without plugin annotation.
```
ljwharbers, line 130: **Nit:** no blank line between this paragraph and "For mutational signatures..." on the next line, so Markdown renders the two as one paragraph.

nextflow_schema.json
AmberVerhasselt: All five were exactly as you described — thanks, I updated `docs/usage.md` and
never came back to the schema. Fixed in `3d324e5`:

- `vep_cadd_snv` / `vep_cadd_indel` — now described as opt-in, in the same shape
  as the existing `vep_eve` text, with the per-session foreign-file cache as the
  reason. The "largest download a default run makes" and "nothing is published to
  the output directory" claims are gone.
- `vep_polyphen_sift_db` — `release-115/`, and it now says why it is pinned.
- `vep_clinvar` — the contig claim is softened to match usage.md
  (synonyms, then adding/stripping `chr`; not guaranteed per contig; annotates
  nothing rather than erroring).
- `skip_vep_plugins` — ~1.4 GB GRCh38 / ~14 GB CHM13, and notes CADD and EVE add
  nothing unless set.

Two more in the same file were stale for a reason you could not have seen from
the diff, both falsified by my own later changes:

- `vep_revel_tbi` / `vep_eve_tbi` called themselves "unused" alongside a release
  zip. Since the index guard in `837efe0` they are *rejected* there, not ignored.
- `vep_revel` / `vep_eve` said to pass a `.zip`. Since `2e7eb85` the test is
  whether the value is already a bgzipped file — which matters most for EVE,
  whose own download URL has no file extension at all.

subworkflows/local/utils_nfcore_lrsomatic_pipeline/tests/main.function.nf.test
AmberVerhasselt: Correct, and this one is worse than a nit on my side: **my reply on the
end-to-end thread claimed these tests run in CI, and that was false when I wrote
it.** Apologies — you were reading the tags more carefully than I was.

`tag "small"` added in `3d324e5`. Confirmed with
`nf-test test --tag small --dry-run`, which now lists "Test VEP plugin resolution
functions"; before the change the selector returned nothing for the file. Full
run is 25 tests in ~126 s locally, so it sits comfortably in the small set.

CITATIONS.md
AmberVerhasselt: Fixed — `release-115/`, matching `conf/igenomes.config` and the source table in
`docs/usage.md`.

docs/usage.md
AmberVerhasselt: Fixed.

workflows/lrsomatic.nf
AmberVerhasselt: Fixed. `validateVepPluginParams()` rejects the combination up front:

```groovy
if (params.vep_custom && !(params.vep_args =~ /--custom file=/)) {
    error("--vep_custom: needs a matching '--custom file=...' entry in --vep_args ...")
}
```

The error names the param, says why, and gives a copy-pasteable `--vep_args` with
a placeholder `--custom file=` entry to substitute into. Two tests cover it — one
for the rejection, one asserting that a `--vep_args` carrying its own
`--custom file=` is accepted.

I went with your validation suggestion rather than reshaping the ClinVar entry.
Reordering to `--custom short_name=ClinVar,file=...` would also dodge the
`/--custom file=[^,\s]+/` regex, but it depends on VEP treating those keys as
order-independent, and an explicit error beats a silent near-miss that hinges on
argument order.

CHANGELOG.md
AmberVerhasselt: Both fixed. `#192` → `#193` in link text and URL.

On redistribution — right that it was stale, and it was wrong in two directions
once prepared files started being published. The entry now says what is actually
hosted: the AlphaMissense GRCh38 `.tbi` and the CHM13 protein-space table, both
derived from AlphaMissense and so CC BY 4.0, with attribution already in
`CITATIONS.md`. The matching sentence in `docs/usage.md` is rewritten the same
way.

Worth separating the two ideas, since the old sentence conflated them: publishing
to `--outdir` is not redistribution — it copies a file the user downloaded
themselves onto their own disk. But `<outdir>/vep_plugins/` does sit inside the
tree people tar up for collaborators, and a 667 MB non-commercial REVEL table
riding along in a shared archive *would* be. So `docs/usage.md` carries an
`[!IMPORTANT]` telling people to exclude that directory when sharing results.

bin/prepare_vep_plugin_data.sh
AmberVerhasselt: My earlier reply said switching the trap to `EXIT` was harmless. It was not, and
the first real run caught it — worth a follow-up on that thread:

> Correction to the above: the `EXIT` trap as I first wrote it broke the script.
> `work` is declared `local`, so a `RETURN` trap sees it but an `EXIT` trap —
> which fires after the function has returned — does not, and under `set -u` that
> is a fatal "work: unbound variable". `VEPPLUGIN_REVEL` did all its work, wrote
> both output files, and then exited 1. Fixed in `2e7eb85` by expanding the path
> when the trap is defined rather than when it fires.
>
> The same run also surfaced a pre-existing one: `find … | head -n1` under
> `set -o pipefail` returns 141 when the producer is still writing. With EVE's
> ~3000 per-protein VCFs that fired every time and took the run down;
> `prep_revel` only survived the identical construct because `find` matches a
> single file there and finishes before `head` exits. Both now use `mapfile`.

robert-a-forsyth (approved):

workflows/lrsomatic.nf
AmberVerhasselt: # VEP plugins never reached VEP — cause, fix, and verification

## The bug

Both cluster test runs of this branch finished `COMPLETED 0:0` having annotated
nothing with plugins. `grep -c -- "--plugin"` on the VEP command Nextflow actually
submitted was **0** on both assemblies:

| Run | Assembly | Elapsed | `--plugin` in the VEP command |
|---|---|---|---|
| 11573859 | CHM13/T2T | 4 h 17 m | 0 |
| 11574609 | GRCh38 | 4 h 37 m | 0 |

The GRCh38 run did the full plugin preparation first — `WGET_REVEL` → `UNZIP_REVEL`
→ `VEPPLUGIN_REVEL`, the same for EVE, plus staging AlphaMissense and both CADD
tables — and then invoked VEP without a single `--plugin`. All of that work was
discarded, and the run reported success.

**Cause.** `workflows/lrsomatic.nf` set `params.vep_plugin_args = vep_plugins.args`,
and `conf/modules.config` read it back inside an `ext.args` closure. A param
assigned in the workflow body is not visible to `conf/modules.config` closures, so
the read found nothing and `?: ''` turned that into an empty string. Nextflow said
so once per VEP task — 8× on CHM13, 7× on GRCh38:

```
WARN: Access to undefined parameter `vep_plugin_args`
```

The `?: ''` is what made it silent: a missing value became a plugin-free VEP run
instead of an error.

That the split is about *where the read happens* is shown by `params.vep_genome`,
assigned the same way ten lines earlier, equally undeclared, which warns not at all
and arrives correctly as `--assembly GRCh38` — because it is read in the workflow
body and passed as a module input.

Note that declaring `vep_plugin_args` in `nextflow.config` is **not** a fix, even
though Nextflow's warning suggests exactly that. It would silence the warning and
leave VEP plugin-free.

## The fix

Pass the argument string to the module directly instead of routing it through
`params`, the same way the ASCAT patch in this repo passes `val(genomeVersion)`.

- `modules/nf-core/ensemblvep/vep/main.nf` — new `val plugin_args` input, emitted
  after `${args}` in the `vep` call. `ENSEMBLVEP_VEP` is already patched here, so
  this extends an existing patch; regenerated `ensemblvep-vep.diff` accordingly.
- `conf/modules.config` — `SOMATIC_VEP` and `GERMLINE_VEP` drop the phantom param,
  becoming `ext.args = { "$params.vep_args" }`, identical to the existing `SV_VEP`
  block.
- `workflows/lrsomatic.nf` — removed the `params.vep_plugin_args` assignment and
  passed `vep_plugins.args` to `GERMLINE_VEP` and `SOMATIC_VEP`, `''` to `SV_VEP`.

`grep -rn vep_plugin_args` across the repo now returns nothing.

## Verification 1 — the arguments are rendered

Every VEP task from the post-fix runs, counting occurrences in the submitted
`.command.sh`. Before the fix every row here was `0 / 0 / (none)`.

| Assembly | Task | `--plugin` | `--custom` | Resources |
|---|---|---:|---:|---|
| GRCh38 | DLBCL1_SOMATIC_VEP | 4 | 1 | AlphaMissense, CADD, EVE, REVEL + ClinVar |
| GRCh38 | DLBCL1_GERMLINE_VEP | 4 | 1 | AlphaMissense, CADD, EVE, REVEL + ClinVar |
| GRCh38 | BL1_GERMLINE_VEP | 4 | 1 | AlphaMissense, CADD, EVE, REVEL + ClinVar |
| GRCh38 | DLBCL1_SV_VEP | 0 | 0 | (none) — intended |
| GRCh38 | BL1_SV_VEP | 0 | 0 | (none) — intended |
| CHM13 | DLBCL1_SOMATIC_VEP | 2 | 1 | AlphaMissenseProtein, PolyPhen_SIFT + ClinVar |
| CHM13 | BL1_SOMATIC_VEP | 2 | 1 | AlphaMissenseProtein, PolyPhen_SIFT + ClinVar |
| CHM13 | DLBCL1_GERMLINE_VEP | 2 | 1 | AlphaMissenseProtein, PolyPhen_SIFT + ClinVar |
| CHM13 | BL1_GERMLINE_VEP | 2 | 1 | AlphaMissenseProtein, PolyPhen_SIFT + ClinVar |
| CHM13 | DLBCL1_SV_VEP | 0 | 0 | (none) — intended |
| CHM13 | BL1_SV_VEP | 0 | 0 | (none) — intended |

Per annotated VCF: **5 resources on GRCh38**, **3 on CHM13**. The difference is
correct — REVEL, EVE and CADD are GRCh38-only, and CHM13 uses the protein-space
`AlphaMissenseProtein` and `PolyPhen_SIFT` instead, which is what
`validateVepPluginParams()` enforces. On GRCh38 SIFT and PolyPhen come from the VEP
cache via `--everything`, so the plugin is not needed there.

Three details worth noting:

- `SV_VEP` is 0 on both assemblies and both samples — the deliberate exclusion
  ("missense and splice scores are meaningless on SEVERUS breakends") survived.
- Germline and somatic are both covered, so all three call sites were fixed.
- CHM13 additionally carries `--dir_plugins .`, which the bundled
  `AlphaMissenseProtein.pm` needs.

The rendered CHM13 command:

```
vep -i somatic_smallvariants.vcf.gz -o BL1_SOMATIC_VEP.vcf.gz \
    --everything --filter_common --per_gene --total_length --offline --format vcf --vcf \
    --dir_plugins . \
    --plugin AlphaMissenseProtein,file=alphamissense_protein_v2023_uniprot-2026_03.tsv.gz \
    --plugin PolyPhen_SIFT,db=homo_sapiens_pangenome_PolyPhen_SIFT_20240502.db \
    --custom file=clinvar_20240624_GCA_009914755.4.vcf.gz,short_name=ClinVar,format=vcf,type=exact,coords=0,fields=CLNSIG%CLNREVSTAT%CLNDN \
    ...
```

and GRCh38:

```
    --plugin AlphaMissense,file=AlphaMissense_hg38.tsv.gz \
    --custom file=clinvar_20260829.vcf.gz,short_name=ClinVar,...,fields=CLNSIG%CLNREVSTAT%CLNDN \
    --plugin CADD,snv=whole_genome_SNVs.tsv.gz,indels=gnomad.genomes.r4.0.indel.tsv.gz \
    --plugin REVEL,file=revel_grch38.tsv.gz \
    --plugin EVE,file=eve_merged.vcf.gz \
```

Each file is referenced by its staged basename, and `revel_grch38.tsv.gz` /
`eve_merged.vcf.gz` are the names the prep tasks write — so the argument side and
the `extra_files` staging side agree.

## Verification 2 — the annotations actually land in the VCF

Rendering the arguments is not the same as annotating, so here are populated field
counts from the output VCFs themselves.

**GRCh38, same sample and same file, before and after.** Both runs parse the
identical 43,630 CSQ entries, so this is a like-for-like comparison:

| Field | Before | After |
|---|---:|---:|
| am_class, am_pathogenicity | 0 | 132 |
| CADD_PHRED, CADD_RAW | 0 | 42,946 |
| REVEL | 0 | 121 |
| EVE_CLASS, EVE_SCORE | 0 | 19 |
| ClinVar | 0 | 96 |
| ClinVar_CLNSIG, ClinVar_CLNDN | 0 | 90 |
| *SIFT (from the VEP cache)* | *150* | *150* |
| *PolyPhen (from the VEP cache)* | *135* | *135* |

The last two rows are the control: SIFT and PolyPhen come from the cache rather
than a plugin, and they are **unchanged at 150 and 135**. VEP was always working —
it simply had no plugins. Every plugin-derived field went from empty to populated,
and nothing else moved.

CADD's 42,946 of 43,630 is expected: unlike the missense-only predictors, CADD
scores variants of every consequence class.

**CHM13** — somatic VCF, 51,909 CSQ entries:

| Field | Populated |
|---|---:|
| AlphaMissenseProtein_match | 263 |
| am_class, am_pathogenicity | 206 |
| SIFT_pred, SIFT_score | 246 |
| PolyPhen_humvar_pred, PolyPhen_humvar_score | 217 |
| ClinVar, ClinVar_CLNSIG, ClinVar_CLNDN | 163 |



The ClinVar `fields=CLNSIG%CLNREVSTAT%CLNDN` list survives its unusual
`%`-separated syntax, and `AlphaMissenseProtein_match` distinguishes `gene_aa` from
`aa_mismatch`, so the protein-space lookup is behaving as designed.

ljwharbers (approved):

PR #196, opened 2026-09-17
Ship lrsomatic_report as a container instead of vendoring its source

ljwharbers:
## Why

`assets/lrsomatic_report` was a hand-maintained copy of the report tool — 38 files, ~9,900 lines — because `nextflow run IntGenomicsLab/lrsomatic` does not fetch submodules. Every release of that tool meant re-copying six paths and bumping a version string in **seven** places that nothing cross-checked. Its own `VENDORED.md` said so:

> Nothing checks these for consistency; miss one and the snapshots go stale silently.

And PR CI never ran the report module test — `nf-test.yml` filters pull requests to `tags: small` — so a bad vendor landed green.

The tool now ships inside its own image and the copy is deleted. **Updating it is a tag bump on two lines**, which is how `MODKIT_PILEUP` and the three `SIGPROFILER_*` modules already work here.

A submodule would not have fixed this: it still would not reach users, and it does nothing about the dependency half of the problem.

## What changed

**The version is no longer hardcoded.** lrsomatic_report v1.5.0 adds `--version`, so the versions topic reads it out of the container with an `eval` and cannot disagree with the image it ran in. The command is `env -u R_HOME render_report.R --version`: Apptainer forwards the host environment, and R prints `WARNING: ignoring environment value of R_HOME` on *stdout* when `R_HOME` is set, which would otherwise land in the captured version string. Docker does not forward it, so CI cannot see this — it is reproducible on the cluster and is covered by the module run linked below, which exports `R_HOME` deliberately.

> ⚠️ This renumbers `nxf_out_eval_N` for processes declared *after* `LRSOMATICREPORT`, so it invalidates `-resume` for them. The module sits near the end of the workflow, so the blast radius is small — but don't expect a fast resume on existing work dirs.

**Builtin gene panels moved to `assets/gene_lists/`** and reach the tool through its new `--gene-lists-dir`. They had to go somewhere: launch-time validation of `--report_gene_panel` globs that directory *on the host*, which cannot see inside an image. Staging them as ordinary user panels would not have been equivalent — they'd take the user-panel branch and be renamed `lymphoid-custom` against the image's own copy — hence the upstream flag.

The pipeline now depends on nothing inside the container except `render_report.R` being on `PATH`, and adding a panel is a pipeline change rather than a tool release. **User-supplied panel TSVs are completely unaffected.**

**The report tool is pinned at v1.6.0**, which carries the v1.4.0 VEP-plugin annotation work (AlphaMissense, ClinVar, CADD, REVEL, EVE, and SIFT/PolyPhen from the `PolyPhen_SIFT` plugin on CHM13), the v1.5.1 fix for the SV table's stale `panel_hit` cell, and v1.6.0's optional `applies_to` panel column plus a footer that reads the tool's `VERSION` at render time.

**The builtin `lymphoid` panel is rebuilt** and scoped per table using that column: 72 genes filtering both tables becomes 234 rows — 128 with per-reference spans as the SV set, 148 symbols as the coding-mutation set, 42 on both. Ticking `lymphoid` therefore filters the two tables on different gene sets now. This is a behaviour change for anyone who used it before, and it is written up in `docs/usage.md` as well as the changelog.

**Conda is dropped from the nf-test matrix.** The module loses its `conda` directive and gains the deepvariant/modkit-style guard, because an `environment.yml` would install the R/Quarto stack without `render_report.R`. `conf/test.config` does not set `skip_report`, so the report runs in all five pipeline tests and the `conda` shard would fail on every `dev`→`main` PR. The alternative — `skip_report` in `test.config` — would have cost the report its CI coverage under docker and singularity too, and required rewriting five snapshots.

> Worth stating plainly: conda previously ran only on `dev`→`main` PRs, and now runs nowhere. That removes conda coverage for **every** module, not just this one, until `lrsomatic-report` reaches bioconda. A conda regression elsewhere in the pipeline will now land green.

**Also:** the script block loses its `activate.d` sourcing loop (the image handles that now); `--report_src` is gone, since it pointed at source that no longer exists here; and `.gitattributes` drops its `linguist-vendored` line, which leaves it identical to the nf-core template, so its `files_unchanged` exemption goes with it.

## Panel corrections found in review

The `lymphoid` pair is re-synced from the report tool's repository, and four rows were repaired here rather than copied verbatim. `assets/gene_lists/README.md` records them so a future re-sync re-applies them; they are worth fixing upstream too.

- **`EWSR1`, `KDM6B` and `SIK3` carried their CHM13 coordinates in the hg38 file.** They were the only three of 128 coordinate rows whose hg38 and t2t cells were byte-identical, off by **+463 kb**, −96 kb and +15 kb against GENCODE v46. All three are `sv`-scoped and therefore matched on position alone, so an hg38 `EWSR1` rearrangement — a classic lymphoma translocation partner — sat well outside the 100 kb SV-span rule and was filtered *out* of a `lymphoid`-ticked SV table, while unrelated events near chr22:29.73 Mb were filtered in.
- **`PRKBC` was a transposition of `PRKCB`.** On an `snv` row with no coordinates, symbol matching is the only mechanism there is, so the row matched nothing and `PRKCB` — present in the panel before the rebuild — was silently absent.
- **`RCK` is an obsolete alias for `DDX6`** and already carried DDX6's coordinates; renamed, so `panel_hit` names a symbol VEP output can actually carry.
- **`FAM46C` was dropped** as a dead duplicate of `TENT5C`, which is already in the list. That is what takes the count from 235 rows to 234.

## Testing

- **`nf-test modules/local/lrsomaticreport` on Slurm, apptainer, against the 1.6.0 image** — the real check, since PR CI will not run it (`tags: small`). Output in a comment below.
- **A full report renders end to end**, run directly against the same fixture the module test uses, with `HOME`/`TMPDIR` redirected the way a Nextflow task has them. `--gene-lists-dir` replaces the bundled panel set rather than merging with it, which is what this pipeline relies on.
- **`Sys.glob` follows the staged symlink**, checked against `assets/gene_lists` through a symlinked directory. That was the one real risk in staging the panels as a directory.
- **Panel integrity**, re-derived rather than trusted: hg38 and t2t agree gene-for-gene and cell-for-cell on `applies_to`; no duplicate symbols, no `start > end`, no contigs outside `chr1`–`chr22`/`X`/`Y`; coordinates present on exactly the 128 SV-capable rows and absent on all 106 `snv`-only rows; every coordinate row cross-checked against GENCODE v46, and `sarcoma` verified at 140/140 symbols with no discrepancy over 20 kb.
- **Upstream R suite: 380 passing, 0 failures**, covering `--gene-lists-dir`, `--version`, `VERSION` vs the conda recipe, and that the entrypoint parses.
- `nextflow lint` on the changed files: 0 errors. `nextflow config -profile test` parses with no `report_src`.
- `pre-commit run --all-files`: passes.

Not run: `nf-core pipelines lint` crashes in this environment fetching the `nf-core/modules` cache (`Branch 'master' not found`), before any check executes. Unrelated to this change — CI covers it.

## Known gap, unchanged by this PR

`test_SOMATIC_VEP.vcf.gz` declares only `SIFT` and `PolyPhen` in its `CSQ` header and has no `##VEP=` line, so every plugin-annotation path added in report v1.4.0 renders its "absent" branch and stays untested. Worth a fixture whose `CSQ` header names the plugin fields plus `##VEP=… dbSNP="156" COSMIC="99"` — separate piece of work.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


AmberVerhasselt:
## Verdict

**Approve with minor changes.** The design is sound and the mechanics are right. I pulled the
Apptainer image on wICE and rendered a report end to end — it works, and the PR's central claim
(that the panel set now belongs to the pipeline, not the tool) holds. Nothing here blocks merge.

One runtime issue is worth fixing while the module is open. It is pre-existing rather than
introduced by this PR, but it is a guaranteed failure for us on wICE and Mindwell.

---

## 🟡 Should fix — `XDG_CACHE_HOME` is not sandboxed, and quarto crashes on our cluster

`modules/local/lrsomaticreport/main.nf:50-53` sandboxes `HOME`, `TMPDIR`, `TMP` and `TEMP` into
the task directory, on the stated reasoning that "Quarto/Deno write under `$HOME` and `$TMPDIR`,
which clusters may mount read-only". It does not sandbox `XDG_CACHE_HOME` — and Deno prefers
that over `$HOME/.cache`.

Apptainer/Singularity inherit the host environment, so on any site where `XDG_CACHE_HOME` points
somewhere Nextflow does not bind into the container, the render dies:

```
error: Uncaught (in promise) Error: Read-only file system (os error 30):
       mkdir '/data/leuven/365/vsc36520/.cache/quarto'
    at quartoCacheDir (file:///opt/conda/bin/quarto.js:33225:12)
```

Reproduced on KU Leuven VSC, where `XDG_CACHE_HOME=$VSC_DATA/.cache` is set by default. The host
directory is perfectly writable — it simply isn't bind-mounted, so it presents as read-only
inside the container. Note the R-side error is unhelpful here: `quarto`'s own error formatter
hits a `glue()` bug and masks the message above behind `Error running quarto CLI from R`.

One line fixes it, and the render then succeeds (verified — 2.2 MB HTML produced):

```groovy
export HOME=\$PWD
export TMPDIR=\$PWD/tmp TMP=\$PWD/tmp TEMP=\$PWD/tmp XDG_CACHE_HOME=\$PWD/.cache
mkdir -p "\$TMPDIR"
```

`XDG_DATA_HOME` and `XDG_CONFIG_HOME` are unset in our environment and fall back to `$HOME`,
which is already sandboxed, so they need nothing.

**To be clear: this is not a regression from #196.** Those `export` lines are unchanged context
in the diff, and the old Wave image had the same exposure. Raising it here because the module is
already open and the cost is one line.

## 🟡 Should fix — `meta.yml` version key breaks the house convention

`modules/local/lrsomaticreport/meta.yml` keys the third version element `"versions"` in both the
`output:` and `topics:` blocks. Every other eval-based module in the repo keys it on the literal
command instead — e.g. `modules/local/sigprofiler/assignment/meta.yml:87`:

```yaml
- 'python -c ''import importlib.metadata as m; print(m.version("SigProfilerAssignment"))''':
    type: string
    description: The command used to generate the version of the tool
```

Suggest `- "render_report.R --version":` with
`description: The command used to generate the version of the tool`, in both blocks.

## 🟡 Consider — the conda guard fires hours late

The guard at `main.nf:31-34` is correct and matches `deepsomatic/*`, but it only trips when the
task is dispatched — after alignment, calling, VEP and ASCAT have already run. A conda user
loses a full pipeline run to a message telling them to pass `--skip_report`.

`subworkflows/local/utils_nfcore_lrsomatic_pipeline/main.nf` already validates at launch in
`validateReportGenePanels()`. The same profile check there — error unless `params.skip_report` —
would turn this into a two-second failure. This is a repo-wide pattern, so fine to defer.

## 🟢 Nits

- `CHANGELOG.md`: a blank line sits between the third and fourth `#196` entries, making that a
  loose list, inconsistent with the rest of the section. The v1.4.0 entry also ends
  `. (@ljwharbers).` — stray space before the paren, unlike every other entry.
- The PR says a version bump is now "two lines". It is four: `main.nf:11`, `main.nf:12`, and two
  hardcoded `1.5.0` tags in the new `docs/usage.md` prose. Consider dropping the explicit tag
  from usage.md so the docs cannot drift from the module.
- Dropping `conda` from the whole nf-test matrix costs the entire pipeline its conda coverage
  because of one module. Keeping conda in the matrix with `--skip_report` would preserve it. The
  inline comment does state the restore condition, so this is defensible as-is.

---

## What I verified

Run on wICE, apptainer 1.4.4, no registry credentials configured.

- **The images pull anonymously.** `apptainer pull --disable-cache
  oras://ghcr.io/ljwharbers/lrsomatic-report-sif:1.5.0` → 723 MB SIF; the OCI image is public too.
  The container selector, and both `autoMounts` settings, match the in-repo convention.
- **`render_report.R --version` prints exactly `1.5.0`**, so the `eval()` output matches all six
  updated snapshots. `--gene-lists-dir` exists at v1.5.0 as used.
- **Dropping the `activate.d` sourcing is safe** — the riskiest change here, and it holds up.
  `/opt/conda/bin/render_report.R` in the image is a wrapper that sources the conda activation
  hooks itself before `exec Rscript …/share/lrsomatic_report/bin/render_report.R`, so
  `QUARTO_SHARE_PATH` and friends resolve without any caller-side dance.
- **End-to-end render works** with the module's own `test_SOMATIC_VEP.vcf.gz` (once
  `XDG_CACHE_HOME` is handled): 2.2 MB HTML, panels embedded, published path and MultiQC version
  entry match `docs/output.md`. Dropping a throwaway TSV into the staged `gene_lists/` made it
  appear in the report and be accepted as a `--gene-panel` name, which proves `--gene-lists-dir`
  really does override the container's bundled set.
- **Clean removal:** no residual `report_src` / `assets/lrsomatic_report` / `VENDORED` references
  outside CHANGELOG history; `.gitattributes` now matches the nf-core template byte-for-byte, so
  dropping its `files_unchanged` exemption is right; the prettier/pre-commit exclusions were
  removed consistently.

**Not verified:** the new per-column claims in `docs/output.md` (AlphaMissense, ClinVar, CADD,
REVEL, EVE columns; `PolyPhen_SIFT` on CHM13; the "Annotation sources" footnote) — my test VCF
carries no plugin fields. They match the upstream v1.4.0 release notes, and the three
`- real render` nf-tests cover them.

ljwharbers:
Thanks @AmberVerhasselt — pulling the image and rendering end to end on wICE was more than I asked for, and the `XDG_CACHE_HOME` find is a good catch. Pushed as 469e3a3.

## 🟡 `XDG_CACHE_HOME` — fixed

You're right, and it's worse than "some sites": this pipeline's own cluster sets it. `XDG_CACHE_HOME=$VSC_DATA/.cache` on Mindwell too, so every apptainer run of this module here was exposed the same way yours was on wICE.

Taken exactly as you wrote it:

```groovy
export TMPDIR=\$PWD/tmp TMP=\$PWD/tmp TEMP=\$PWD/tmp XDG_CACHE_HOME=\$PWD/.cache
```

I widened the comment above it so the next reader knows *why* the cache dir is in that list (Deno prefers it over `$HOME/.cache`; apptainer forwards the host value but doesn't bind the path), and added a `Fixed` changelog entry crediting you with the root cause — including the `Error running quarto CLI from R` masking, which is the part that would have cost someone a day.

Noted that it's pre-existing rather than a regression from this PR. Fixing it here anyway, for the reason you gave.

For the record, `LRSOMATICREPORT` is the only process in the repo that exports `HOME` at all (`grep -rl "export HOME=" modules/ subworkflows/ workflows/ bin/`), so no sibling module needs the same treatment.

## 🟡 `meta.yml` version key — fixed

Changed in both the `output:` and `topics:` blocks:

```yaml
- "render_report.R --version":
    type: string
    description: The command used to generate the version of the tool
```

Confirmed the convention holds across every eval-based module here, not just `sigprofiler/assignment` — `vcfsplit`, `cramino`, `metaextract` and `fibertoolsrs/*` all key on the literal command with that exact description.

## 🟡 Conda guard timing — not changing, deliberately

Fair description of the cost, but conda is not a supported way to run this pipeline, so a conda user losing a run to a late error isn't a case worth engineering around. The dispatch-time guard stays as-is and consistent with the other nine conda-guarded processes (`sigprofiler/*`, `deepsomatic/*`, `deepvariant/*`, `modkit/pileup`). If that ever changes, it should change for all of them at once rather than for the report alone.

## 🟢 Nits — all three applied

- Blank line removed; that `#196` entry is a tight list item again.
- `...panel hero cards. (@ljwharbers).` → `...panel hero cards (@ljwharbers).`
- You were right that it was four lines, not two. Dropped both hardcoded tags from `docs/usage.md` and pointed the prose at the module instead, so `grep -rn "1\.5\.0" docs/ modules/local/ conf/ nextflow.config` now returns only `main.nf:11` and `main.nf:12`. The claim in the PR body is true as of this commit.

On the conda matrix: leaving it, as you suggest. The inline comment carries the restore condition and the bioconda recipe is the real fix.

## Module nf-test — run, and it passes

The output the PR body promised. PR CI still won't run these (`nf-test.yml` filters pull requests to `tags: small`), so here it is. Run on Mindwell, `dedicated_big_bigmem`, 8 CPU / 32 GB, apptainer 1.4.4, `--profile debug,test,apptainer`, at 469e3a3:

```
host XDG_CACHE_HOME (the thing being guarded against): /data/leuven/364/vsc36452/.cache

Test Process LRSOMATICREPORT

  Test [a28b6180] 'no optional inputs - stub'
    > Pulling Apptainer image oras://ghcr.io/ljwharbers/lrsomatic-report-sif:1.5.0
    PASSED (196.165s)

  Test [bca35630] 'vep somatic vcf - real render'
    PASSED (13.936s)

  Test [30c738b2] 'custom gene panel tsv - real render'
    PASSED (13.805s)

  Test [919586ad] 'several gene panels - real render'
    PASSED (13.695s)

SUCCESS: Executed 4 tests in 237.607s

git status of module snapshot after run:
```

Two things fall out of that:

- **The snapshot needs no regeneration.** It ran *without* `--update-snapshot` and `git status` on `main.nf.test.snap` came back empty, so the hand-edited `1.5.0` values agree with what `render_report.R --version` actually prints. That closes the last outstanding item in the PR description.
- **It is also the regression test for the fix above.** The three real renders ran on a host that sets `XDG_CACHE_HOME` outside the bound work tree — the same exposure that failed for you — and they render. I didn't run them unpatched, so I'm relying on your reproduction for the failing half.

The 196s on the first test is the `oras://` pull; the image caches after that.

Also worth recording for anyone reading this later: the remaining "Merge order" steps in the PR description are already done — [lrsomatic_report#24](https://github.com/ljwharbers/lrsomatic_report/pull/24) is merged, `v1.5.0` is tagged, and both packages are public (your anonymous pull proves it).

Agreed on the untested plugin-annotation paths, and on the fixture that would close them — a `CSQ` header naming the plugin fields plus a `##VEP=` line. Separate work; better as its own issue than grown into this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ljwharbers:
Bumped this branch to **lrsomatic_report 1.5.1** (both image references, the module snapshot
and the five pipeline snapshots).

1.5.0 renders the SV table's `panel_hit` column once per row node — DataTables only renders a
cell when it creates the `<tr>`, and the table is `deferRender` + `Scroller`, so the node is
reused for every later draw. Ticking a gene panel filtered the rows correctly but left every
already-drawn row with the empty cell it was born with: on a T2T DLBCL report, a breakend
filtered in on `MYC` 139 kb away sat there with no reason shown, three rows above a correctly
labelled `RHOA (B, 650 kb)`. 1.5.1 invalidates the table in `redraw()`, so the cells, the
column's search box and its sort all follow the selection
(ljwharbers/lrsomatic_report#26).

Nothing else moves: the version reaches the snapshots through `render_report.R --version`,
and the report HTML is snapshotted by filename only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ljwharbers:
## report v1.6.0 is out, and it needs one thing here that a tag bump alone won't cover

[ljwharbers/lrsomatic_report#28](https://github.com/ljwharbers/lrsomatic_report/pull/28) is merged and [`v1.6.0`](https://github.com/ljwharbers/lrsomatic_report/releases/tag/v1.6.0) is released. Both images are published and anonymously pullable (checked — `HTTP 200` on each manifest without credentials, so the visibility step from the merge order above is already done):

```
ghcr.io/ljwharbers/lrsomatic-report:1.6.0
oras://ghcr.io/ljwharbers/lrsomatic-report-sif:1.6.0
```

### What it changes

A panel TSV can now scope each gene to one table via an optional `applies_to` column — blank/`both` for both tables, `snv` for small variants only, `sv` for SVs only. Coordinates are consequently required only on rows that can match an SV. A panel **without** the column behaves exactly as before.

`lymphoid` is rebuilt on top of that: 235 genes merged from an NHL gene bed (128 with per-reference spans, the SV list) and the Twist list (149 symbols, the small-variant list), so `IGH`/`IGK`/`TRB`/`TRA/D` filter SVs only and `MYD88`/`NOTCH1` filter small variants only. `sarcoma` is untouched — no `applies_to` column, all 140 genes still match both tables.

### Why that lands on this PR specifically

> Builtin gene panels moved to `assets/gene_lists/` … **replaced, not merged, which is exactly what this pipeline relies on** … adding a panel is a pipeline change rather than a tool release.

That's the crux. Because `--gene-lists-dir` replaces the bundled set, **bumping the container alone changes nothing for users** — this branch's `assets/gene_lists/lymphoid.{hg38,t2t}.tsv` are still the old 72-gene files, and they are what the tool will read. The new panel only arrives if those two TSVs are copied across as well.

### ⚠️ Order matters, and the wrong order fails silently

If the new TSVs land while the container is still `1.5.1`, the v1.5.1 loader rejects the 107 `snv`-scoped rows for having blank coordinates — and `load_all_gene_panels()` catches that as a `message()`, not an error. Run against the real v1.5.1 loader:

```
Skipping gene panel 'lymphoid': ... has missing or non-numeric coordinates for:
ACTG1, ANKRD26, ARID1B, B2M, BCOR (and 102 more)
panels offered: sarcoma
lymphoid present: FALSE
```

The report renders perfectly well, with `lymphoid` simply absent from the chip row. Nothing fails, no snapshot moves. **So the container bump and the TSV swap want to be the same commit** — or, if this PR is close to merging, do the whole thing as a follow-up rather than splitting it.

### Checklist

- [ ] `modules/local/lrsomaticreport/main.nf` — `1.5.1` → `1.6.0` on the two container lines
- [ ] `assets/gene_lists/lymphoid.{hg38,t2t}.tsv` — replace with the v1.6.0 files
- [ ] `assets/gene_lists/README.md` — carries the panel schema, so it needs the `applies_to` section
- [ ] `modules/local/lrsomaticreport/tests/main.nf.test.snap` — pins `"1.5.1"` in both `versions_lrsomaticreport` blocks; `--update-snapshot`
- [ ] Changelog, alongside the existing #196 entries

The `eval('render_report.R --version')` change in this PR is doing its job here: the container tag is the only place the version is written by hand, so there's no `meta.yml` string to keep in step.

### One behavioural note worth a second opinion

The 107 Twist-only genes are `snv`-scoped, so **a deletion that removes `MYD88` or `NOTCH1` entirely will not show up in the panel-filtered SV table.** That was the explicit request upstream and it's a one-cell edit per gene to change, but it's the kind of thing that reads as a filter bug months later if nobody flagged it.

### Upstream verification

R suite **435 passing, 0 failures** (was 380), covering the scope split, the conditional coordinate requirement, the unrecognised-value error, the all-`snv` panel reporting `has_coords = FALSE`, and `sarcoma`'s no-column-means-both contract. Full renders on t2t (`--gene-panel lymphoid --gene-panel sarcoma`) and hg38: the payload carries 149 SNV symbols against 128 SV symbols and 128 intervals for `lymphoid`, 140/140/140 for `sarcoma`, and each reference gets its own coordinates (t2t `MYC` 128,863,685 vs hg38 127,735,434). SV count unchanged against the pre-change baseline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ljwharbers (commented):
Full review of the combined diff (`dev...51c918c`), covering the containerisation, the 1.6.0 bump and the rebuilt `lymphoid` panel. Everything below that needed fixing is already fixed in 51c918c; the rest is recorded so it is not re-litigated.

## Fixed in this branch

**1. Three hg38 panel rows carried CHM13 coordinates.** `EWSR1`, `KDM6B` and `SIK3` were the only three of 128 coordinate rows whose hg38 and t2t cells were byte-identical — every other gene differs between the two files, and `sarcoma` has no such row. They were the T2T values:

| gene | was (hg38 file) | GENCODE v46 hg38 | delta |
| --- | --- | --- | --- |
| `EWSR1` | chr22:29731500-29763945 | chr22:29268009-29300525 | **+463,491** |
| `KDM6B` | chr17:7738322-7760122 | chr17:7834217-7854796 | −95,895 |
| `SIK3` | chr11:116858511-117113708 | chr11:116843402-117098437 | +15,109 |

All three are `sv`-scoped, so they are matched *only* on position. On a GRCh38 run an `EWSR1` rearrangement — a classic lymphoma translocation partner — sat 463 kb from the panel interval, outside the 100 kb SV-span rule entirely, so a real `EWSR1` SV was filtered **out** of a `lymphoid`-ticked SV table while unrelated events near chr22:29.73 Mb were filtered in. `KDM6B` was 96 kb off, right at the boundary. Replaced with the GENCODE v46 spans; the t2t file was correct and is untouched.

**2. `PRKBC` was a transposition of `PRKCB`.** Not a symbol in GENCODE v46. It is `snv`-scoped with no coordinates, so symbol matching is the only mechanism it has — the row matched nothing and `PRKCB`, which *was* in the panel before the rebuild, went silently missing. Fixed.

**3. `RCK` → `DDX6`, and `FAM46C` dropped.** `RCK` is an obsolete alias for `DDX6` and already carried DDX6's coordinates, so it still matched positionally but named a symbol no VEP output can emit. `FAM46C` is the former name of `TENT5C`, which is already in the list — a dead duplicate. Dropping it takes the panel from 235 rows to 234. These four repairs are now recorded in `assets/gene_lists/README.md` so a future re-sync re-applies them; **they are worth fixing in `lrsomatic_report` too**, since that is where the source lists live.

**4. The version `eval` captured an R warning on Apptainer.** `eval('render_report.R --version')` captures stdout, and R prints `WARNING: ignoring environment value of R_HOME` **on stdout** whenever `R_HOME` is set. Apptainer forwards the host environment, so on any host with `R_HOME` exported — the default here, since R comes from micromamba — the versions topic recorded:

```
WARNING: ignoring environment value of R_HOME
1.6.0
```

and every one of the six snapshots would have failed. Docker does not forward host env, which is exactly why CI is green on this. Now `eval('env -u R_HOME render_report.R --version')`, verified in the container (`1.6.0`, clean) and by the nf-test run below, which exports `R_HOME` deliberately.

**5. `docs/usage.md` documented a format the shipped panel violates.** It stated the coordinate rule as "all three or none" and never mentioned `applies_to`, while the builtin `lymphoid` now carries that column and leaves coordinates blank on 106 rows. It now documents the column, the column-vs-row distinction for coordinates, and — separately — the `lymphoid` behaviour change, which previously appeared only in the changelog: 106 coding genes including `MYD88` and `NOTCH1` no longer match SVs at all, so a whole-gene deletion of `MYD88` does not reach a `lymphoid`-filtered SV table, and 17 genes of the old panel are gone.

## Verification

`nf-test modules/local/lrsomaticreport --profile +singularity` on Slurm with `R_HOME` exported, against `oras://ghcr.io/ljwharbers/lrsomatic-report-sif:1.6.0`:

```
Test Process LRSOMATICREPORT
  'no optional inputs - stub'            PASSED (5.953s)
  'vep somatic vcf - real render'        PASSED (16.612s)
  'custom gene panel tsv - real render'  PASSED (14.078s)
  'several gene panels - real render'    PASSED (14.158s)

SUCCESS: Executed 4 tests in 50.807s
```

The six snapshots were hand-edited rather than regenerated, so this run is what confirms they are right: the versions entry matched `1.6.0` with the fix in place. PR CI does not run these (`tags: small`).

Also checked: `nextflow lint` 0 errors; `pre-commit run --all-files` all hooks pass; both GHCR packages are public and carry `1.6.0` (anonymous manifest fetch returns 200).

Panel integrity re-derived rather than trusted: hg38 and t2t agree gene-for-gene and cell-for-cell on `applies_to`; no duplicate symbols, no `start > end`, no contigs outside `chr1`–`chr22`/`X`/`Y`, no CRLF; coordinates present on exactly the 128 SV-capable rows and absent on all 106 `snv`-only rows; every remaining coordinate row cross-checked against GENCODE v46 (`DLEU2`, `KDM2B`, `NBEAP1`, `TP63` differ by 20–68 kb on one edge only, which is ordinary transcript-set variation between the bed and GENCODE, not a lifted coordinate); `sarcoma` verified at 140/140 symbols with no discrepancy over 20 kb. The immunoglobulin and TCR loci (`IGH`, `IGK`, `IGL`, `TRA/D`, `TRB`, `TRG`) are deliberately not GENCODE gene features and are matched positionally.

## Not a defect, but worth stating

**Dropping `conda` removes conda coverage for the whole pipeline, not just this module.** The reasoning in the diff is sound and documented — `conf/test.config` does not set `skip_report`, so the report runs in all five pipeline tests and the conda shard would fail on every `dev`→`main` PR. But conda previously ran only on `dev`→`main` PRs, and now runs nowhere, so a conda regression anywhere else in the pipeline will land green until `lrsomatic-report` reaches bioconda. Now called out in the PR description.

## Checked and clean

- The container directive, conda guard and `oras://` fallback match the `MODKIT_PILEUP` / `SIGPROFILER_*` pattern exactly, including `task.ext.singularity_pull_docker_container`.
- The 86 six-field `sv` rows against a seven-field header are safe — `load_gene_panel` reads with `fread(..., fill = TRUE)`.
- `README.md` inside the staged `gene_lists/` is harmless: the tool globs `*.tsv`, and `reportBuiltinGenePanels()` filters on `.tsv` too.
- `${projectDir}/assets/gene_lists` resolves correctly from both `workflows/lrsomatic.nf:1188` and the subworkflow helper; the launch-time `--report_gene_panel` validation globs names only and never parses cells, so `applies_to` cannot affect it.
- `XDG_CACHE_HOME=$PWD/.cache` is not separately `mkdir`-ed, but with `HOME=$PWD` it resolves to the path Quarto already created — not a regression.
- The module test's sentinel genes `SMAD4` and `VHL` are still absent from both builtins, so the rebuilt panel does not undermine what that assertion proves.
- No dangling `report_src` / `assets/lrsomatic_report` / `VENDORED.md` references outside historical changelog entries; `.gitattributes` now matches the nf-core template exactly, so dropping its `files_unchanged` exemption is correct.
- The `#plugin-fields-in-the-csq-annotation` anchor added in `docs/output.md` resolves to the real heading.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

AmberVerhasselt:
## Verdict

**Approve.** Both findings from my first pass are fixed, correctly. The two things added on top
since then — an `R_HOME` fix on the version eval, and a rebuilt `lymphoid` panel — both hold up
under testing. One cosmetic inaccuracy remains, and one note about how the PR is scoped.

Everything below was run on wICE with apptainer 1.4.4 against the real 1.6.0 image and this PR
head's actual `assets/gene_lists/`.

---

## Round-1 findings: both fixed

- **`XDG_CACHE_HOME`** — `main.nf` now exports `XDG_CACHE_HOME=$PWD/.cache` alongside `HOME` and
  `TMPDIR`, with an accurate comment, and the `Fixed` changelog entry names the VSC symptom and
  the masked R-side error. This was the one that made the render a guaranteed failure for us.
- **`meta.yml` version key** — now keyed on the literal command in both the `output:` and
  `topics:` blocks, matching `sigprofiler/*`.

Both round-1 nits are gone too: the CHANGELOG loose-list blank line, and the hardcoded tag in
`docs/usage.md` (it now points at the module).

## The new `R_HOME` fix is real — I reproduced the bug

Worth saying explicitly, since it is subtle and CI cannot see it. With `R_HOME` exported into the
container — which Apptainer does from the host environment — `render_report.R --version` emits on
**stdout**:

```
WARNING: ignoring environment value of R_HOME
1.6.0
```

That whole two-line string would have gone into `versions.yml` and broken the snapshot.
`env -u R_HOME` gives exactly `1.6.0`. Confirmed the warning is on stdout rather than stderr, so
`eval()` really would have captured it, and confirmed on both the 1.5.0 and 1.6.0 images. Docker
does not forward `R_HOME`, which is why this only ever showed up under Apptainer.

## The rebuilt `lymphoid` panel checks out

This is the biggest change since round 1 — 72 genes to 234 rows with a new `applies_to` column —
so I verified it against the data rather than taking the changelog's word:

- **Structure:** 234 rows in each of hg38/t2t, identical gene sets, no duplicate symbols, no
  `start > end`. `applies_to` splits 42 blank / 86 `sv` / 106 `snv`, so the SV set is 128 and the
  both-lists overlap is 42 — matching the claims exactly.
- **The coordinate rule holds:** all 106 `snv` rows leave coordinates blank, and *zero*
  blank-or-`sv` rows are missing or partial. So the documented "coordinates required only on rows
  that can match an SV" is actually true of the data.
- **Every gene-level claim is accurate:** the old panel was exactly 72 genes; exactly 27 of them
  are now `snv`-only; exactly 17 are dropped, and all six named examples are in that set; the
  rearrangement partners (`IGH`, `IGK`, `IGL`, `TRA/D`, `TRB`, `TRG`, `DUSP22`) are all new and
  `sv`-scoped; the seven claimed `sv`-only genes are all `sv`.
- **The four corrections are real:** `PRKBC`→`PRKCB`, `RCK`→`DDX6`, `FAM46C` dropped with
  `TENT5C` present, and the three contaminated hg38 rows shifted by **+463 kb / −96 kb / +15 kb**
  — exactly the offsets stated. No gene shares coordinates between the two reference files any
  more, which was the tell for those rows in the first place.

**Scoping works end to end**, not just on paper: in the rendered report, the `lymphoid` symbol
set used for small-variant matching holds exactly 148 genes (snv + blank) with **zero** `sv`-only
genes leaking into it, and `MYD88`/`NOTCH1` present as expected.

**A bad `applies_to` value is a hard error**, as `README.md` promises — injecting `typo` aborts
with `unrecognised applies_to value(s): 'typo' (YTHDF2)`, naming the file, the value and the gene,
and produces no output.

**The ragged rows are harmless.** 86 of the `sv` rows omit the trailing empty `notes` field (6
fields against a 7-column header), which strict TSV readers reject. The loader uses
`fread(..., fill = TRUE)` and skips the comment block by counted offset, and the render confirms
it: no error, no misparsed row.

## Also verified

`env -u R_HOME render_report.R --version` returns exactly `1.6.0`, matching all six updated
snapshots. A full render against this PR's `assets/gene_lists/` succeeds (2.2 MB HTML), with and
without `--gene-panel lymphoid`. The report footer now reads `v1.6.0` — my round-1 render with
the 1.5.0 image showed `v1.4.0`, which is exactly the hardcoded-footer bug 1.6.0 replaces with a
`VERSION` read.

---

## 🟢 One cosmetic thing

The changelog gives the small-variant source list as "149 symbols from `twist_genes.tsv`", so the
entry reads 128 + 149 − 42 = **235** against its own stated 234 rows. The resulting SNV set is
**148** — confirmed both in the TSV and in the rendered report's symbol set — because `FAM46C`
was dropped as a `TENT5C` duplicate. The changelog does mention that drop, just not next to the
arithmetic, and `assets/gene_lists/README.md` spells it out ("takes the count from 235 rows to
234"). Suggest "149 symbols (148 after dropping the `FAM46C` duplicate)" so the numbers reconcile
on the page.

## 🔵 A note on scope, not a defect

A PR titled "ship the report as a container" now also changes what `--report_gene_panel lymphoid`
returns — a real output change for anyone using the default panel, bundled into an infrastructure
change. It is documented loudly (a blockquote in `docs/usage.md`, a long CHANGELOG entry), so
nothing is hidden. Raising it only because it makes the PR harder to bisect if the panel rebuild
later turns out to need revisiting.

AmberVerhasselt (approved):

PR #197, opened 2026-09-17
Run ClairS-TO's Verdict against CHM13 instead of GRCh38 only

ljwharbers:
## What this does

ClairS-TO's Verdict module tags tumour-only calls as germline, somatic or subclonal somatic from copy number and tumour purity, which it derives from ASCAT loci, allele and GC content files. Those files only describe the assembly they were built on, and ClairS-TO resolved them under hardcoded GRCh38 names — so on CHM13, germline variants went untagged and leaked into `somatic.vcf.gz`. That is the source of the CHM13 tumour-only germline leak (~10k high-VAF germline SNVs on DLBCL1, shifting the signature profile towards SBS40a).

With `--genome CHM13` the pipeline now builds a CHM13 Verdict resource set and hands it to ClairS-TO as `--cna_resource_dir`.

## How

`CLAIRSTO` moves to `ghcr.io/ljwharbers/clairs-to:0.5.1-verdict-chm13` (fork of HKU-BAL/ClairS-TO), which resolves those resources from `--cna_resource_dir` and checks the loci against `--ref_fn`.

No new hosting was needed: the CHM13 ASCAT loci/allele/GC zips are already igenomes attributes, and their internal filenames are already the ones the resolver expects. A new `CLAIRSTO_CNA_RESOURCES` module lays them out; only the GC file is renamed.

```
cna_resources/
  loci_files/G1000_loci_CHM13_chr{1..22,X}.txt
  allele_files/G1000_alleles_CHM13_chr{1..22,X}.txt
  GC_G1000_CHM13.txt      # renamed from GCcontent_SNPloci_CHM13.txt
                          # no RT_*.txt -> LogR corrected for GC content only
```

CHM13 has no `ascat_loci_rt` attribute, so GC-only correction falls out by construction. That is what ClairS-TO recommends for CHM13: no replication timing file is published for it, lifting the GRCh38 one over leaves purity identical and ploidy within 0.001, and it is poorly anchored across the acrocentric contigs.

The three unzips move out from behind the `skip_ascat` gate since ClairS-TO now wants the same files; `UNZIP_RT` stays behind it. `--clairsto_cna_resources` takes a prepared directory instead, validated at startup for the layout and for links reaching outside it — only the directory itself is staged into the task, so such links would be missing inside the container and Verdict would be quietly disabled, which is the failure this PR exists to prevent.

## ⚠️ GRCh38 results change too

Besides 0.4.2 → 0.5.1, the fork fixes Verdict's GC window selection: upstream read the 25bp window's correlations with the other windows rather than with LogR, so the correction window was chosen by the wrong quantity. GRCh38 Verdict tags will move. This was a deliberate choice — one maintained version rather than a genome-conditional container.

## Also in the ClairS-TO module

- `--conda_prefix` and the SIF are now selected under `-profile apptainer`, not only `-profile singularity`.
- `--sample_name` is no longer passed twice (it was in both the module and `ext.args`).
- 0.5.1 renames its output VCFs after `--sample_name`; they are renamed back, so the published names, `VCFSPLIT` and the snapshot file listings are unchanged. The prefix cannot be forced back via `--snv_output_prefix`: ClairS-TO rewrites that flag whenever its value is still the default.

## Verified

- **New module**: 4 nf-tests (layout, GC rename, GC name preserved when already `GC_*`, ambiguous-GC failure). Tagged `small`, so it runs on PRs.
- **Layout against the tool's own resolver**: built the set from the real ASCAT CHM13 zips and ran the image's `resolve_cna_resources` / `loci_beyond_reference`. No errors, prefixes discovered, `rt_file: None` (GC-only), nothing missing across all 23 Verdict contigs.
- **Mismatch guard**: the same set is accepted against a CHM13 reference and rejected against GRCh38 on 17 contigs, so a misconfiguration costs the tags with a warning rather than producing wrong ones.
- **Pipeline snapshots**: default, clair_only, consensus, deep_only, union regenerated with `-profile debug,test,apptainer`. The only change is the recorded `clairsto` version in the four snapshots with a tumour-only sample — no file listing or md5 moved.
- **GRCh38 unchanged**: with the flag omitted, the logs confirm ClairS-TO still uses its own bundled GRCh38 set including `RT_G1000_hg38.txt`.
- `nf-core pipelines lint`: 427 passed, 0 failed. `nextflow lint` and `pre-commit` clean.

## Not covered here

CI has no CHM13 test data, so the end-to-end CHM13 run is manual. The purity gap first noted here (Verdict 0.42 / 2.87 against R ASCAT 0.50 / 2.02 on DLBCL16) is resolved below: it was a bug in Verdict's ASCAT port, and the tags now come from ASCAT's estimate.

## Update: Verdict tags now come from ASCAT's purity and copy number

The purity discrepancy flagged under *Not covered here* has been investigated (`VERDICT_PURITY_FINDINGS.md` in the validation run directory) and turned out to be bugs in Verdict's Python port of ASCAT, not a fit ambiguity: the segmentation's noise estimate lacked R's `mad()` constant of 1.4826, so almost no segment was recognised as allelically balanced and whole genomes were fitted as ploidy ~3 at too low a purity; copy number segments were merged across chromosome boundaries, which left 5-15% of PASS variants in no segment and therefore never evaluated for tagging; and optima were labelled one grid step off. All three are fixed in the fork (`ljwharbers/ClairS-TO@c0687e8`), and the image pin moves to `0.5.1-verdict-chm13-c0687e8`. Ploidy now lands at 1.95-2.02 on the three CHM13 genomes (R ASCAT: 1.96-2.02); purity is still up to 0.14 from R's and depends on the PCF penalty.

So the pipeline no longer relies on Verdict's own estimate when it does not have to:

- **`CLAIRSTO_VERDICT_TAG`** (new, `modules/local/clairsto/verdict_tag`) applies ClairS-TO's own Verdict tagging script to ASCAT's `purityploidy.txt` and `segments.txt`. A `bin/ascat_to_verdict.py` converter writes the two tables Verdict reads, respelling ASCAT's `1..22,X` to the VCF's contig names and refusing to hand Verdict a purity of 0 when ASCAT found no solution (the calls then pass through untagged, as when purity is above 0.6). Output names are unchanged (`snv.vcf.gz`, `indel.vcf.gz`), so `VCFSPLIT`, publishing and the snapshot file listings are untouched; the tables the tags were computed from are published beside them.
- **`CLAIRSTO`** runs with `--disable_verdict` when ASCAT is in the run (`conf/modules.config`), and its untagged VCFs stay unpublished.
- **`ASCAT`** moves before small variant calling. It only ever needed the BAMs; it was fed from `severus_input`, which made it wait for phased germline calls it does not use.
- **`--skip_ascat`** keeps today's behaviour: Verdict tags from its own estimate inside ClairS-TO.

Verified: 4 nf-tests for the new module against the pinned image (tagging at purity 0.5 with the expected germline / somatic / subclonal calls and contig respelling; pass-through at purity 0.7; pass-through with no purity written when ASCAT has no solution; stub), `nextflow lint` and `pre-commit` clean. The pipeline tests run with `skip_ascat = true`, so they exercise the fallback path and their snapshots do not move; the ASCAT path was exercised with a stub run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


conf/modules.config
```diff
@@ -504,14 +504,32 @@ process {
         ]
     }
     withName: '.*:CLAIRSTO' {
-        ext.args = { "--sample_name ${meta.id}" }
+        // --sample_name is passed by the module itself, from ext.prefix.
+        // With ASCAT in the run, Verdict's own purity/CNA estimate is not used: CLAIRSTO_VERDICT_TAG
+        // redoes the germline tagging from ASCAT's tables and publishes the tagged VCFs under the
+        // same names, so these untagged ones stay unpublished.
+        ext.args   = { params.skip_ascat ? '' : '--disable_verdict' }
```
ljwharbers, line 511: `--disable_verdict` travels in `ext.args`, which is the one knob users are expected to override. A `-c` config with `withName: '.*:CLAIRSTO' { ext.args = '--snv_min_af 0.1' }` silently drops the flag while ASCAT is still in the run, so ClairS-TO spends hours computing its own Verdict tags and then `CLAIRSTO_VERDICT_TAG` appends a second set on top of them: the same record ends up with both `Verdict_Somatic` and `Verdict_Germline` in INFO, and `VCFSPLIT` keys on exactly those.

Worth driving from the module instead of the overridable string, e.g. a `val disable_verdict` input fed from the same condition the subworkflow branches on, so the two cannot disagree.

subworkflows/local/utils_nfcore_lrsomatic_pipeline/main.nf
```diff
@@ -642,3 +635,55 @@ def methodsDescriptionText(mqc_methods_yaml) {
...
+            error("--clairsto_cna_resources: '${resource_dir}' has no ${sub}/ sub-directory. Expected layout: loci_files/<prefix>chr1.txt ..., allele_files/<prefix>chr1.txt ..., GC_<name>.txt, and optionally RT_<name>.txt.")
+        }
+        def first_contig = sub_dir.listFiles().findAll { entry -> entry.name.endsWith('chr1.txt') && entry.name.size() > 'chr1.txt'.size() }
+        if (first_contig.size() != 1) {
+            error("--clairsto_cna_resources: ${sub}/ holds ${first_contig.size()} files ending in 'chr1.txt'; ClairS-TO derives the per-contig prefix from exactly one. Keep one resource set per directory.")
+        }
+    }
+
+    def gc_files = dir.listFiles().findAll { entry -> entry.name.startsWith('GC_') && entry.name.endsWith('.txt') }
+    if (gc_files.size() != 1) {
```
ljwharbers, line 663: The layout check covers `GC_*.txt` and the `chr1` file per sub-directory, but not `RT_*.txt` and not the other 22 contigs. Both gaps land in the failure mode this function exists to prevent: ClairS-TO's `_single_candidate` rejects two `RT_*.txt` (easy to end up with after a `.bak`), `missing_resource_files` rejects an incomplete set, and either way it sets `disable_verdict` with only a log warning, so the run completes with germline calls leaking into `somatic.vcf.gz`.

The same 0-or-1 test applied to `RT_*.txt`, plus a check that all 23 `<prefix>chr{1..22,X}.txt` are present, would close it.

ljwharbers (commented):
Reviewed the CHM13 Verdict work and the ASCAT-driven tagging that was added on top. The design holds up: I checked `paired_data` is set on both rows of a pair and false for tumour-only, so the `ascat_tumoronly_ch` filter and the `failOnMismatch` join cannot see a paired sample; that moving ASCAT earlier changes none of its inputs, only drops a wait on the phased VCF it never used; that the launcher skips `check_cna_resources` entirely under `--disable_verdict`; and that `tag_germline_variant.py` ignores the `Sample` column, returns without writing above purity 0.6, and indexes its own output. The nf-core ASCAT module really does write the two column layouts `bin/ascat_to_verdict.py` reads.

I pushed three commits for the things worth fixing, described in each message. Summary:

- **The CNA resource set was built where nothing reads it.** `build_clairsto_cna` ignored `skip_ascat`, so on the default CHM13 path the directory was assembled and staged into a `CLAIRSTO` that runs `--disable_verdict`. The docs then told users to confirm the set from a `VERDICT CNA RESOURCE DIRECTORY` log line that only appears when Verdict runs inside ClairS-TO. Gated on `--skip_ascat`, with a warning when `--clairsto_cna_resources` is passed without it.
- **A missing ASCAT attribute failed silently.** With a custom CHM13 genome entry lacking `ascat_loci` or `ascat_alleles`, `PREPARE_REFERENCE_FILES` emits an empty channel, the three-way join never fires, `.first()` never emits, and `CLAIRSTO` plus every tumour-only output vanishes from a run that reports success. Now an error at startup.
- **The two published tables did not exist on a `--skip_ascat` run**, though `docs/output.md` listed them for it. ClairS-TO leaves its own pair in its work directory under the tool's default sample name, so they are lifted out under the names the other path publishes.
- **A compression failure shipped untagged calls.** `bgzip` and `tabix` run unchecked inside ClairS-TO, so a failure left no `.vcf.gz` and was indistinguishable from "purity above the cutoff, nothing tagged": the module copied the untagged VCF over it and exited 0, putting the germline leak back. A leftover plain VCF is that signature, and now fails the task.
- **Explicit output prefixes**, so the four renames are fixed names rather than a glob. A sample called `SAMPLE` got no rename at all and failed.
- **A stub nf-test for `TUMORONLY_SMALLVAR`** covering both tagging paths, ~12 s, tagged `small`. A full-pipeline stub test would be the wider net but is not currently possible: `UNZIP`'s stub leaves an empty directory so the ASCAT reference channels never emit, and `METAEXTRACT` declares `env` outputs its stub never sets.

The rest are comments rather than changes, inline below. Nothing here blocks merge.

One correction worth recording: the `ascat_loci_gc` URL for CHM13 looks like it should be switched to `raw.githubusercontent.com` for consistency with its two siblings, and it must not be. That file is stored in Git LFS, so the raw URL returns a 134-byte pointer while the redirecting `github.com/.../raw/` form resolves to the real 138 MB archive. The re-download on every `-resume` is a consequence of the redirect, and fixing it needs content-addressed staging, not a URL swap.

subworkflows/local/utils_nfcore_lrsomatic_pipeline/main.nf
```diff
@@ -642,3 +635,55 @@ def methodsDescriptionText(mqc_methods_yaml) {
...
+        error("--clairsto_cna_resources: '${resource_dir}' holds ${gc_files.size()} files matching GC_*.txt; ClairS-TO needs exactly one.")
+    }
+
+    def real_root = dir.toRealPath()
```
ljwharbers, line 667: `toRealPath()` and `Files.isSymbolicLink` assume the default filesystem, while the schema advertises `format: directory-path`, which nf-schema happily accepts for `s3://`. On a remote path this throws `UnsupportedOperationException` and the user gets a stack trace instead of the message this function was written to give them.

Either refuse a non-local scheme with a clear error, or skip the symlink audit when the path is not on the default filesystem.

bin/ascat_to_verdict.py
```diff
@@ -0,0 +1,95 @@
...
+    with open(args.segments) as fp, open(args.cna_out, 'w') as out:
+        header = fp.readline().rstrip('\n').split('\t')
+        col = {name: header.index(name) for name in ('chr', 'startpos', 'endpos', 'nMajor', 'nMinor')}
+        out.write('Sample\tChromosome\tStartPosition\tEndPosition\tnMajor\tnMinor\n')
+        for line in fp:
+            fields = line.rstrip('\n').split('\t')
+            chrom = spell_like_vcf(fields[col['chr']], contigs)
+            if contigs and chrom not in contigs:
+                unmatched.add(chrom)
+            out.write('\t'.join([args.sample, chrom, fields[col['startpos']], fields[col['endpos']],
```
ljwharbers, line 77: The four numeric columns are copied through as text, but `tag_germline_variant.py` parses them with `int()`. R's `write.table` has no `scipen` set, so a double-typed coordinate column can come out as `2e+07`, and a segment with undetermined copy number as `NA`. Either kills the tagging task for that sample, and takes its whole tumour-only branch with it.

`int(float(x))` for the coordinates and dropping `NA` copy-number rows here, where the message can name the line, would be more forgiving than failing inside Verdict.

modules/local/clairsto/cna_resources/environment.yml
```diff
@@ -0,0 +1,8 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - conda-forge::coreutils=9.5
```
ljwharbers, line 7: The script's real workhorse is `find` (findutils), which is not declared here. It works today because the wave image and any realistic conda environment carry it, but the declaration should say so.

ljwharbers:
## Full-sample validation of the ASCAT-based tagging, on three CHM13 tumour-only genomes

Re-ran the three CHM13 samples on this branch after merging `origin/dev` (head `500e9c0`), full
pipeline, 5 h 26 m, completed successfully. This exercises `CLAIRSTO_VERDICT_TAG` on real data —
`conf/test.config` sets `skip_ascat = true`, so CI never reaches this path outside the module's own
nf-test.

ASCAT's `purityploidy.txt` and `segments.txt` came out byte-identical to the previous run, so the
only thing that changed between the two tag sets is which purity was believed.

| sample | ASCAT purity / ploidy | segments | `Verdict_Germline`: 0.4.2 → Verdict's own → ASCAT's |
| --- | --- | ---: | --- |
| `DLBCL16_oldpb` | 0.50 / 2.0188 | 96 | 0 → 4,532 → **12,376** |
| `FL11_newpb` | 0.70 / 1.9763 | 49 | 0 → 7,276 → **0** |
| `BL1_ont` | 0.46 / 1.9615 | 96 | 0 → 11,473 → **13,872** |

DLBCL16 also gets 6,324 `Verdict_Somatic` and 40 `Verdict_SubclonalSomatic`. The two tagged samples
gain germline calls, which is the expected direction — ASCAT reads both genomes as near 1/1 diploid
at a higher purity than Verdict's own port did, so more variants sit at a germline-consistent
allele fraction.

### Worth being explicit about: this is a trade, not a free win

FL11's ASCAT purity is 0.70, above Verdict's 0.6 tagging threshold, so it now gets **no tags at
all** and its somatic set returns to where ClairS-TO 0.4.2 left it:

| | 0.4.2 | Verdict's own | ASCAT's |
| --- | ---: | ---: | ---: |
| somatic calls | 30,054 | 22,328 | 30,028 |
| PASS SNVs at AF ≥ 0.45 | 16,440 | 9,259 | 16,442 |

That AF ≥ 0.45 band is where germline leakage sits, so for this sample the germline separation the
PR exists to provide is switched off. Whether that is right depends on a purity question this PR
does not settle: R says 0.70, the fixed Verdict port says 0.56, Wakhan says 0.68. Wakhan siding
with R is the only independent evidence and it favours leaving FL11 untagged — which is why the PR
goes this way — but it is worth a reviewer knowing that one of three samples loses tagging.

### Checks

- **Pass-through is lossless.** FL11 goes in and comes out at 3,577,360 records on both snv and
  indel; DLBCL16 at 3,550,439. The untagged copy drops nothing.
- **No segment defects.** No `*_Tumor_CNA.txt` row has start > end, and each has 23 contigs with no
  cross-chromosome span.
- `--disable_verdict` reached `CLAIRSTO` on all three samples, and `CLAIRSTO_VERDICT_TAG` published
  `<sample>_Tumor_Purity_Ploidy.txt` / `_Tumor_CNA.txt` next to the VCFs.
- 19 CACHED, 118 COMPLETED, 13 FAILED-then-retried. All 13 are the same cluster I/O artefact —
  Nextflow could not read `.exitcode` off GPFS within the 600 s timeout and retried; every retry
  succeeded.

### Two notes for the merge

- `origin/dev` merged with one conflict, in `CHANGELOG.md`, and only because both sides appended to
  `Changed`; both sets of entries are kept. The four shared `tests/*.nf.test.snap` files differed
  only in the `clairsto` and `lrsomatic_report` version strings, so they merged correctly without
  regeneration.
- #196 replaced a `val('1.3.2')` with an `eval(...)` in `modules/local/lrsomaticreport/main.nf`.
  Nextflow numbers `eval` outputs with a global counter in include order and writes that number into
  `.command.sh`, so everything included after `LRSOMATICREPORT` (`workflows/lrsomatic.nf:35`) loses
  its resume cache. Same mechanism as #193. Not a defect, but it makes `-resume` across this merge
  much less effective than it looks.

robert-a-forsyth (approved):

